PR #204, opened 2026-04-19
fix(local): nginx ingress class, web memory limit, and build cache

vsuran:
- Add className: nginx to local ingress so Docker Desktop's nginx controller picks up the Ingress resource without manual annotation
- Raise web pod memory limit 1Gi → 2Gi to prevent OOMKilled on Next.js cold starts under Tilt
- Mount an emptyDir for .next build cache so the web container doesn't compete with the read-only filesystem
- Add cache_from to Tiltfile docker_build so Tilt reuses the locally-built image layer cache instead of re-pulling on each run
- Replace hardcoded Auth0 dev credentials with placeholder values; developers should set their own in a local override or .env

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -74,6 +74,7 @@ docker_build(
     'terrapod-api',
     context='.',
     dockerfile='docker/Dockerfile.api',
+    cache_from=['terrapod-api:latest'],
     live_update=[
         sync('./services/terrapod', '/app/terrapod'),
         sync('./alembic', '/app/alembic'),
--- a/helm/terrapod/values-local.yaml
+++ b/helm/terrapod/values-local.yaml
@@ -54,8 +54,8 @@ api:
         default_provider: "auth0"
         oidc:
           - name: "auth0"
-            issuer_url: "https://dev-zqwrse0xlal03qiu.us.auth0.com/"
-            client_id: "AymR3SnEXQAdKE0t5NUPezczpcgpcXK5"
+            issuer_url: "https://dev-1234.us.auth0.com/"
+            client_id: "abc123"
             scopes:
               - openid
               - profile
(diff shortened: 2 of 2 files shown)


mattrobinsonsre:
Thanks for the contribution, @vsuran. Closing this one, but here's where the pieces landed and what I'd like to do differently going forward.

## What's being adopted

Three of the five changes were scope-clean local-dev improvements. I've cherry-picked them into #214:

- `cache_from` on the API `docker_build`
- Web memory limit `1Gi → 2Gi`
- Dedicated `emptyDir` for the Next.js `.next` cache

Credit noted in the PR description.

## What isn't landing as-is, and why

- **`ingress.className: nginx`** — Docker Desktop's nginx controller needs this, but Rancher Desktop (the maintainer's setup) ships Traefik by default; forcing `nginx` as the class means the Ingress gets claimed by nothing and local dev breaks. A hardcoded class value isn't portable across local-dev stacks.
- **Replacing the Auth0 `issuer_url` and `client_id` with placeholders** — those values aren't secrets (the client secret lives in a K8s Secret, not in `values-local.yaml`), and committing the working dev tenant is the "clone and go" onboarding path for anyone who wants to use SSO locally. Stripping them to `dev-1234` / `abc123` without a documented alternative regresses that.

### Suggested workaround for both

Use a per-developer override file that isn't committed, and layer it on top of `values-local.yaml`:

```zsh
# my-values.yaml (gitignored, not committed)
ingress:
  className: nginx   # your Docker Desktop setup
api:
  config:
    auth:
      sso:
        oidc:
          - name: "auth0"
            issuer_url: "https://<your-tenant>.us.auth0.com/"
            client_id: "<your-client-id>"
```

Then point Tilt at both:

```zsh
tilt up --port 10352 -- -f helm/terrapod/values-local.yaml -f my-values.yaml
```

This way the committed `values-local.yaml` stays working for everyone, and you get your Docker-Desktop-specific overrides without touching shared config.

## Process request for next time

Please open an issue before opening a PR of this size. A short "here's the setup I'm on and the friction I'm hitting" issue gives us a chance to discuss alternatives (like the override pattern above) before you spend time implementing something that turns out to conflict with the maintainer's setup. Happy to design something together once the problem is clear.

PR #275, opened 2026-05-08
fix: runner-driven run transitions + listener SSE recovery

### LINKED ISSUE #273: Reconciler doesn't recover in-flight runs after API pod restart; stale-run timeout fails to error them
## Symptom

A speculative plan run gets stuck in \`planning\` indefinitely when:

1. Listener launches the Job and POSTs \`/api/v2/listeners/{lid}/runs/{rid}/job-launched\` (run transitions \`queued → planning\`).
2. Runner completes plan, POSTs \`/api/v2/runs/{rid}/plan-result\` (returns 200) and PUTs \`/api/v2/runs/{rid}/artifacts/plan-log\` (returns 204).
3. K8s job-controller marks the Job \`Completed\`.
4. Around the same window, the API Deployment rolls (HPA, redeploy, OOM, anything that replaces a pod).
5. Listener SSE connections drop with \`peer closed connection without sending complete message body (incomplete chunked read)\` and reconnect.
6. Scheduler comes back up on the new pod and re-registers \`run_reconciler\`.
7. **The reconciler picks up brand-new runs from this point onwards but never comes back to the in-flight runs from before the rollover.** They sit in \`planning\` forever.

The 1-hour stale-run timeout that CLAUDE.md says should error these (\"detects stale runs (>1h with no Job status) and errors them\") doesn't fire either — confirmed in production with runs sitting at >6h, still \`planning\`, no errored-out transition.

## Hypothesised root cause

The in-flight breadcrumb the reconciler depends on is the Redis key \`tp:job_status:{run_id}:{phase}\`, written by the \`plan-result\` POST handler with a 120s TTL. If the reconciler doesn't read that key within 120s (because the API pod that holds the periodic-task lock just died and a new one is still spinning up the scheduler), the breadcrumb evaporates.

After the new pod takes over the lock, the reconciler finds the run in \`planning\` with \`job_name\` set and publishes \`check_job_status\` to the pool's listener SSE channel. But by then:

- The K8s Job has often been TTL-cleaned (default \`ttlSecondsAfterFinished\` is short).
- Even if the Job is still around, the listener has no Redis breadcrumb to write back, so the reconciler reads nothing and silently does nothing on subsequent ticks.

There is no durable fallback: the runner already wrote the plan log to object storage (the canonical artifact) and POSTed the plan-result authoritatively, but neither of those is consulted by the reconciler.

## Trace from a real incident

- \`08:11:23.355\` — Listener creates K8s Job
- \`08:11:23.116\` — \`Run transitioned queued→planning\` (run A)
- \`08:11:23.124\` — \`Run transitioned queued→planning\` (run B)
- \`08:11:36.033\` — \`POST /runs/{B}/plan-result\` 200
- \`08:11:36.264\` — \`POST /runs/{A}/plan-result\` 200
- \`08:11:36.268\` — \`PUT /runs/{B}/artifacts/plan-log\` 204
- \`08:11:36.516\` — \`PUT /runs/{A}/artifacts/plan-log\` 204
- \`08:11:38\` — Job-controller marks Jobs Completed
- \`08:16:02\` — Listeners log \`SSE connection failed\` (API pod going away)
- \`08:16:07\` — Listeners reconnect to a different API pod
- \`08:17:20.853\` — New API pod logs \`Registered periodic task: run_reconciler interval=2\`
- \`08:29:34\` — Reconciler logs \`Plan succeeded\` for an unrelated, later run
- \`<no further log entries for runs A or B; they remain in \`planning\` 6h+ later>\`

The other two parallel runs in the same batch (which finished a few seconds earlier, before the rollover) got reconciled fine.

## Suggested fix

Two-pronged. The first is the structural fix; the second is the safety net.

1. **Make \`plan-result\` POST authoritative for state transition.** The runner is the source of truth for \"plan finished\"; it has the exit code, the plan body, and a runner token scoped to that run. The endpoint should transition \`planning → planned\` (or \`planned_and_finished\` for plan-only) directly, not just write a Redis breadcrumb that may evaporate. This removes the 120s TTL race entirely and survives any number of API pod restarts.

   Today the endpoint is purely advisory (Redis-backed); the reconciler is the only writer of the run state machine. That's elegant in steady state but brittle across restarts.

2. **Reconciler fallback when the breadcrumb is gone.** When a run is in \`planning\` with \`job_name\` set and \`tp:job_status:{run_id}:{phase}\` is missing, the reconciler should:
   - Check whether the plan-log artifact has been uploaded to object storage. If yes → plan finished successfully (or at minimum, finished). Transition accordingly.
   - Or: ask the listener to query Job status from K8s, and if the Job is gone (TTL'd), check artifact storage as the tiebreaker.

3. **Fix the 1-hour stale-run errored-out timeout** — currently runs sit indefinitely in \`planning\`. Whatever query/condition is supposed to surface stale runs is either not being scheduled or its filter doesn't match this case. At minimum, even without the structural fix above, a stuck run should error out after 1h instead of pretending to still be working.

## Workaround in the meantime

Force-cancel the stuck run via the API or DB. Re-trigger the speculative plan by closing+reopening or re-pushing the PR.

## Why this matters

Every API pod redeploy, HPA scale-down, or OOM-kill creates a window where any run that finishes inside it permanently jams. With auto-scaling enabled and \`auto_apply=false\`, this leaves PR checks stuck on \"PENDING\" with no path to resolution short of operator intervention.

mattrobinsonsre:
Closes #273.

## Why

A run can finish its plan/apply Job cleanly in K8s but sit in `planning` (or `applying`) for an hour until the stale-check timeout errors it out. Investigated in the issue: the runner POSTs `/plan-result` successfully, the Job completes, but the reconciler's listener-driven `check_job_status` round-trip silently fails to reach the listener. State machine never transitions; user sees a perpetually pending PR check.

## What

Two interlocking fixes.

### 1. Architectural: runner is now authoritative for transitions

New helpers in `run_service`:

- `complete_plan(db, run, has_changes=None)` — idempotent. Drives `planning → planned` (or `→ applied` for zero-change non-speculative, or `→ confirmed` for auto-apply). Runs the post-plan task-stage gate.
- `complete_apply(db, run)` — idempotent counterpart for `applying → applied`.

Two callers, both routed through these helpers:

- `POST /api/v2/runs/{rid}/plan-result` (existing, was thin) — now calls `complete_plan` directly. Runner has the exit code and diff in hand → its successful POST is sufficient evidence.
- `POST /api/v2/runs/{rid}/apply-result` (**new**) — mirror for the apply phase. `runner-entrypoint.sh` now POSTs this on success.

The reconciler's `_handle_succeeded` is simplified to delegate to the same helpers — same code path, idempotent so the runner-side and listener-side paths are safe to race.

The listener-driven Job-status flow becomes a **fallback** for cases where the runner can't post (OOM-killed before entrypoint runs, network partition, etc.). Stale-check stays as the last-resort safety net.

### 2. Listener SSE recovery

Two robustness changes to `_sse_connect`:

- **Read watchdog** (`TERRAPOD_SSE_READ_TIMEOUT`, default 30s). Every `aiter_lines` read is wrapped in `asyncio.wait_for`. API yields a keepalive every 1s; longer silence = the stream has gone dead even though TCP is nominally open. Timeout → fresh reconnect via the outer loop.
- **Mandatory periodic reconnect** (`TERRAPOD_SSE_MAX_AGE`, default 600s). Even a healthy stream is rebuilt at this cadence so no stale subscription state can survive forever.

Plus visibility: `INFO` log on every dispatched SSE event. Today the only way to tell whether a listener is consuming events is from the absence of output, which is exactly the failure mode we just hit.

## Files

- `services/terrapod/services/run_service.py` — adds `complete_plan` / `complete_apply`.
- `services/terrapod/api/routers/runs.py` — `POST /plan-result` now drives the transition; new `POST /apply-result`.
- `services/terrapod/services/run_reconciler.py` — `_handle_succeeded` simplified to delegate.
- `services/terrapod/runner/listener.py` — SSE watchdog + max-age reconnect + visibility logging.
- `docker/runner-entrypoint.sh` — POSTs `/apply-result` on apply success.
- `services/tests/services/test_run_service.py` — 8 new tests for `complete_plan` / `complete_apply`.

## Test plan

- [x] `make lint` clean
- [x] `make test` — 1237 passing (1229 prior + 8 new)
- [ ] Smoke: deploy to a non-prod cluster, observe a clean run transitions to `planned` in <1s without listener round-trip
- [ ] Behaviour test for the SSE watchdog — block keepalives → observe listener reconnect within `read_timeout`s

## Compatibility

Old runners (no `/apply-result` POST) still work — the reconciler's listener path remains. The `complete_*` helpers' idempotency means upgrading the API ahead of the runner is safe; upgrading the runner ahead of the API is also safe (the new endpoint already exists once API is upgraded).

The new SSE env vars have safe defaults; no Helm chart changes required.

### DIFF
--- a/docker/runner-entrypoint.sh
+++ b/docker/runner-entrypoint.sh
@@ -496,6 +496,15 @@ elif [ "$TP_PHASE" = "apply" ]; then
             EXIT_CODE=1
         fi
     fi
+
+    # Report apply completion to API on success — drives the
+    # `applying → applied` transition without waiting for the listener-
+    # driven Job-status round-trip. Best-effort; the reconciler's listener
+    # path is the fallback.
+    if [ -n "$TP_API_URL" ] && [ -n "$TP_RUN_ID" ] && [ "$EXIT_CODE" = "0" ]; then
+        curl -sSf --max-time 10 -X POST -H "$AUTH_HEADER" \
+            "${TP_API_URL}/api/v2/runs/${TP_RUN_ID}/apply-result" || true
+    fi
 fi
 
 echo "[entrypoint] Phase $TP_PHASE completed with exit code $EXIT_CODE"
--- a/services/terrapod/api/routers/runs.py
+++ b/services/terrapod/api/routers/runs.py
@@ -1185,17 +1185,42 @@ async def report_plan_result(
     body: dict = Body(...),
     db: AsyncSession = Depends(get_db),
 ) -> JSONResponse:
-    """Runner Job reports plan has_changes result.
+    """Runner Job reports plan completion.
 
-    Called by the runner entrypoint after plan completes. The has_changes
-    value is used by the reconciler to set the run's has_changes field.
+    Authoritative state-transition trigger: the runner has the exit code
+    and the diff in hand, so its successful POST here is sufficient
+    evidence that the plan finished. We update `has_changes` and drive the
+    `planning → planned` (or `applied` for zero-change non-speculative)
+    transition directly via `run_service.complete_plan`.
+
+    The reconciler's listener-driven path remains as a fallback for cases
+    where the runner can't post (OOM-killed before the entrypoint runs,
+    network partition, etc.). Both paths land in the same idempotent helper
+    so whichever wins, the second is a no-op.
     """
     run = await _get_run(run_id, db)
 
     has_changes = body.get("has_changes")
-    if has_changes is not None:
-        run.has_changes = has_changes
-        await db.commit()
+    await run_service.complete_plan(db, run, has_changes=has_changes)
+    await db.commit()
+
+    return JSONResponse(content={"status": "ok"})
+
+
+@router.post("/runs/{run_id}/apply-result")
+async def report_apply_result(
+    run_id: str = Path(...),
+    db: AsyncSession = Depends(get_db),
+) -> JSONResponse:
+    """Runner Job reports apply completion.
+
+    Mirror of `plan-result`: authoritative transition trigger driven by the
+    runner's exit. Drives `applying → applied` via `run_service.complete_apply`,
+    which is idempotent against the listener-driven fallback.
+    """
+    run = await _get_run(run_id, db)
+    await run_service.complete_apply(db, run)
+    await db.commit()
 
     return JSONResponse(content={"status": "ok"})
 
--- a/services/terrapod/runner/listener.py
+++ b/services/terrapod/runner/listener.py
@@ -39,6 +39,19 @@ logger = get_logger(__name__)
 # Shutdown flag
 _shutdown = asyncio.Event()
 
+# SSE read watchdog: API sends a keepalive comment every 1s, so any stretch
+# of silence longer than this means the stream has gone silent even though
+# the TCP socket is still open (e.g. API-side Redis pubsub detached, proxy
+# buffering). 30s is generous enough to absorb network jitter without
+# accidentally killing healthy streams.
+_SSE_READ_TIMEOUT = int(os.environ.get("TERRAPOD_SSE_READ_TIMEOUT", "30"))
+
+# Mandatory periodic reconnect — even a healthy connection is torn down
+# and rebuilt at this interval to guarantee no stale subscription state
+# survives forever. 10 minutes balances "cheap to reconnect" against
+# "don't churn unnecessarily."
+_SSE_MAX_AGE = int(os.environ.get("TERRAPOD_SSE_MAX_AGE", "600"))
+
 
 class RunnerListener:
     """Stateless Job launcher — ARC-pattern controller."""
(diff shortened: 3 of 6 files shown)


mattrobinsonsre:
## Local Tilt verification

Ran four end-to-end smoke tests against this branch in Tilt:

| Test | Result |
|---|---|
| Plan-only run (runner-driven transition) | ✓ planning → planned in <1s after `/plan-result` POST. `Plan succeeded` logged from `terrapod.services.run_service` (not `run_reconciler`) — confirms the new path. |
| Auto-apply full flow (plan + apply, both new endpoints) | ✓ queued → applied in 6s. Both `complete_plan` + `complete_apply` fired; `Plan succeeded` and `Apply succeeded` log lines from the new helpers; both `POST /plan-result` and `POST /apply-result` returned 200. |
| Zero-change short-circuit | ✓ planned → applied in 3ms via `complete_planned_as_noop`, no apply Job launched, `has_changes=false` correctly persisted. |
| SSE visibility logging | ✓ every event logged at INFO with `event_type` (was the missing-piece for diagnosis in production). |
| SSE max-age periodic reconnect | ✓ Set `TERRAPOD_SSE_MAX_AGE=30` and observed `"SSE max-age reached, reconnecting"` at exactly the configured age, followed by clean `SSE connecting` → `SSE connected`. Same recovery code path as the read-watchdog. |

Not runtime-tested but reasoned through:

- **Failed plan path** uses `_handle_failed` which is unchanged in this PR.
- **SSE read-watchdog actual stall** — same recovery code path as the max-age reconnect (both return from `_sse_connect` to let the outer loop reconnect). Max-age confirms the path; the watchdog just uses a different trigger.
- **Post-plan task-stage gate** has explicit unit tests (`test_failed_post_plan_stage_errors_run`, `test_pending_post_plan_stage_keeps_run_in_planning`).

### Recommended pre-merge soak

Before promoting to prod, deploy this branch to a non-prod Terrapod (e.g. `dev-eu1`) and let real PR traffic exercise it for 24h. The local Tilt covers the happy paths but real traffic is what surfaces subtle races, retries, and the long-tail edge cases.

Issue #338, opened 2026-05-18
Bug: Runner fails to download binary when version is set without patch number


the3venthoriz0n:
Version: 0.23.4

## Summary:
When a workspace's terraform/tofu version is set to a minor version without the patch number (e.g., 1.11), the runner entrypoint fails to download the binary from both the binary cache and upstream. Setting the full semver (e.g., 1.11.0) works.

## Steps to reproduce:

Set workspace terraform version to 1.11 (or tofu version to 1.11)
Queue a run
Runner fails at download step
Runner log:

```
[entrypoint] Downloading tofu 1.11 (linux/amd64) from binary cache...
curl: (22) The requested URL returned error: 400
[entrypoint] Binary cache unavailable, downloading from upstream...
curl: (22) The requested URL returned error: 404
```

## Expected behavior:
Either:

Resolve 1.11 to the latest patch release (1.11.x) before downloading, OR
Validate at the API level and reject non-semver version strings with a helpful error (e.g., "version must be fully qualified: 1.11.0")

## Actual behavior:
The version string is passed directly to the download URL, resulting in a 404 from upstream since terraform/1.11/ doesn't exist as a release artifact.

## Workaround:
Always set the full semver version (e.g., 1.11.0 instead of 1.11).

As always, thanks you're killing it with these updates!

mattrobinsonsre:
Thanks for the detailed report (and the kind words 🙏). Digging in, there are **two separate things** in your log, and it's worth being straight about both.

**1. The `404` from upstream — a real bug, and we're fixing it.**

You're right that the upstream fallback can't cope with a `x.y` version. The runner entrypoint's "binary cache unavailable, download from upstream" path interpolates the workspace version verbatim into a release-artifact URL (e.g. `.../terraform/1.11/terraform_1.11_linux_amd64.zip`), which only exists for a fully-qualified `x.y.z`. Partial versions like `1.11` *are* designed to be supported — they resolve to the latest matching patch, and this is used extensively in practice — but that resolution currently only happens inside the binary-cache API call, so the upstream fallback never benefits from it. We're fixing this by resolving the version to an exact `x.y.z` server-side at run creation and pinning it on the run, so both the cache path and the upstream fallback always get a concrete version (and an unresolvable version fails fast with a clear message instead of a confusing 404).

**2. The `400` from the binary cache — this is the primary problem, and it's environment-specific.**

[...]

PR #341, opened 2026-05-19
fix: pin S3 SigV4 for presigned URLs + surface storage errors (#339)

### LINKED ISSUE #339: Bug: Presigned S3 URLs fail with KMS encryption — SigV2 incompatible with SSE-KMS
Version: 0.25.0

Environment: AWS GovCloud (us-gov-west-1), EKS, S3 storage backend with SSE-KMS encryption

## Summary
Runner pods fail to download configuration archives and binary cache objects. The API generates presigned URLs using Signature Version 2, but S3 rejects them with InvalidArgument because KMS-encrypted objects require Signature Version 4. The source code documents SigV4 intent but doesn't seem to implement it.

## Error
From inside the runner container, following the presigned URL returned by the API's 302 redirect:

```
<Error>
  <Code>InvalidArgument</Code>
  <Message>Requests specifying Server Side Encryption with AWS KMS managed keys require AWS Signature Version 4.</Message>
  <ArgumentName>Authorization</ArgumentName>
</Error>

```
Runner log (what the operator sees)

```
[entrypoint] Downloading terraform 1.15.3 (linux/amd64) from binary cache...
curl: (22) The requested URL returned error: 400
[entrypoint] Binary cache unavailable, downloading from upstream...
[entrypoint] Downloading configuration...
[entrypoint] No configuration archive (HTTP )
[entrypoint] ERROR: working directory 'terrapod-testing/account-1' not found in config
```

### Root Cause
S3Store._get_client() in services/terrapod/storage/s3.py creates the aioboto3 client without specifying signature_version:

```
self._client = await self._session.client(
    "s3",
    region_name=self._region,
    endpoint_url=self._endpoint_url,
).__aenter__()
```

This causes generate_presigned_url() to produce SigV2 URLs (identifiable by AWSAccessKeyId= and Signature= query parameters). AWS S3 requires SigV4 (X-Amz-Algorithm=AWS4-HMAC-SHA256) for any request involving KMS-encrypted objects.

## Suggested Fix

```
from botocore.config import Config

self._client = await self._session.client(
    "s3",
    region_name=self._region,
    endpoint_url=self._endpoint_url,
    config=Config(signature_version="s3v4"),
).__aenter__()
```

This forces SigV4 for all presigned URLs. SigV4 is backwards-compatible — it works for both KMS-encrypted and non-KMS objects, so there's no downside to always using it.

## How We Diagnosed
Observed runner pods failing with No configuration archive (HTTP ) — no useful error in runner logs
Confirmed API returns 302 with presigned URL via API access logs
Extracted the presigned URL from a runner-debug pod by manually running the tp_curl_download logic step-by-step
Followed the presigned URL with curl -sS (without -f) to get the actual S3 XML error body
Confirmed the URL uses SigV2 format (AWSAccessKeyId=)
Verified that SigV4 presigned URLs generated with Config(signature_version='s3v4') work correctly against the same bucket

## Workaround

Change the S3 bucket encryption from aws:kms to AES256 (S3-managed keys). AES256 does not require SigV4 for presigned URLs, so the SigV2 URLs generated by the current code work correctly.

---
## Secondary Issue: Silent Failure / Poor Error Reporting
The runner entrypoint (runner-entrypoint.sh line 236-237) hides all diagnostic information:

```
tp_curl_download /tmp/config.tar.gz -H "$AUTH_HEADER" \
    "${TP_API_URL}/api/terrapod/v1/runs/${TP_RUN_ID}/artifacts/config" 2>/dev/null || true
2>/dev/null
```
suppresses curl's stderr (where connection errors and HTTP error details appear)
|| true swallows non-zero exit codes

The log message No configuration archive (HTTP ) references an HTTP_CODE variable that is never set by tp_curl_download
The subsequent error working directory not found in config is misleading — it implies a VCS/branch issue when the actual problem is an S3 authentication failure.
The S3 error body (InvalidArgument: Requests specifying Server Side Encryption with AWS KMS managed keys require AWS Signature Version 4) immediately identifies the problem, but it's never surfaced to the operator.

## Recommendation: 

At minimum, remove 2>/dev/null from the config download call and log the response body on failure. Ideally, tp_curl_download should capture and log the HTTP status code and response body when the redirect-follow step fails.

mattrobinsonsre:
Closes #339. Credit to @the3venthoriz0n for an excellent, fully-diagnosed report that pinpointed the file, line, and fix.

## Primary bug — SigV2 presigned URLs

`S3Store._get_client()` created the aioboto3 client with **no `signature_version`**, despite the module docstring claiming *"Presigned URLs use SigV4."* botocore's implicit default varies by region / endpoint / SDK version and can emit a **SigV2** presigned URL (`AWSAccessKeyId=` / `Signature=`). S3 rejects a SigV2 presigned GET for **any SSE-KMS-encrypted object** (`InvalidArgument: Requests specifying Server Side Encryption with AWS KMS managed keys require AWS Signature Version 4`) and rejects SigV2 outright in **GovCloud / newer regions**. Result: every binary-cache and config-archive download `400`s — even with a fully-qualified version.

**Fix:** pin `Config(signature_version="s3v4")` on the client. SigV4 is universally backward-compatible and AWS-recommended; makes the docstring true; deterministic regardless of region/endpoint/SDK default.

## Secondary — why it was undiagnosable

The runner entrypoint masked the real error:
- `tp_curl_download` followed the presigned URL with `curl -sSfL` (`-f` → no body on 4xx).
- The config download added `2>/dev/null || true`, then logged `No configuration archive (HTTP )` against an **unset variable**, and the run limped on to a misleading `working directory '...' not found in config` — pointing the operator at VCS/branch config instead of S3 auth.

**Fix:** `tp_curl_download` now follows the presigned URL **without `-f`**, captures the real HTTP status (`TP_LAST_HTTP`), and logs the storage response body (e.g. the S3 `InvalidArgument` XML) to stderr. The config caller no longer swallows it and reports the accurate status.

## Relationship to #338

This is the root cause of what was actually breaking this operator on **both** tickets — their #338 `400` (even with exact `1.15.3`) was this SigV2/KMS failure, not partial-version resolution. #338's partial-version bug was real-but-separate and is fixed independently in #340. Posting a corrected diagnosis on #338.

## Honesty / scope

The SigV4 fix is verified by a **LocalStack integration test** (a real presigned URL is now `AWS4-HMAC-SHA256`, not `AWSAccessKeyId=`) plus a unit test. It is **not** reproduced against an actual SSE-KMS / GovCloud bucket (none available here) — the reporter independently verified `s3v4` works against their KMS bucket. The residual unknown (*why* botocore emitted SigV2 in their env) does not affect correctness: explicit `s3v4` is right regardless.

## Test plan
- [x] `make lint`, `make test` — **1488 passed** (unit: client pinned to s3v4; LocalStack integration: presigned URL is SigV4)
- [ ] CI green

### DIFF
--- a/docker/runner-entrypoint.sh
+++ b/docker/runner-entrypoint.sh
@@ -147,12 +147,23 @@ trap on_exit EXIT
 # terrapod.local) which may not resolve from inside the cluster. In that case
 # we rewrite the URL to use TP_API_URL (the internal service name) so the
 # download works.
+# Sets TP_LAST_HTTP to the most relevant HTTP status observed. On a
+# storage error (e.g. an SSE-KMS bucket rejecting a SigV2 presigned
+# URL) the storage response body is logged to stderr so the operator
+# sees the real cause instead of a silent failure (#339).
 tp_curl_download() {
     # $1 = output file, remaining args = curl options (URL last)
     _out="$1"; shift
-    # First request: don't follow redirects, capture Location header
-    _headers=$(curl -sSf -D - -o /dev/null "$@" 2>/dev/null)
+    TP_LAST_HTTP=""
+    # First request: don't follow redirects, capture Location header.
+    # No `-f` and no `2>/dev/null` — we want the status and any error
+    # surfaced, not swallowed.
+    if ! _headers=$(curl -sS -D - -o /dev/null "$@"); then
+        echo "[entrypoint] download: initial request failed (network/TLS/DNS)" >&2
+        return 1
+    fi
     _code=$(echo "$_headers" | head -1 | awk '{print $2}')
+    TP_LAST_HTTP="$_code"
     case "$_code" in
         301|302|303|307|308)
             _location=$(echo "$_headers" | grep -i '^location:' | sed 's/^[Ll]ocation:[[:space:]]*//' | tr -d '\r')
@@ -175,15 +186,34 @@ tp_curl_download() {
                     esac
                 fi
             fi
-            # Follow any further redirects (e.g. S3 region/path-style redirects)
-            curl -sSfL -o "$_out" "$_location"
+            # Follow any further redirects (e.g. S3 region/path-style
+            # redirects). NO `-f`: on a 4xx/5xx we want the response
+            # body (e.g. the S3 InvalidArgument XML) written out so we
+            # can show it, not a bare "curl: (22) ... 400".
+            _final=$(curl -sSL -o "$_out" -w '%{http_code}' "$_location" 2>/dev/null || echo 000)
+            TP_LAST_HTTP="$_final"
+            case "$_final" in
+                2*) : ;;  # success
+                *)
+                    echo "[entrypoint] download: storage returned HTTP $_final for the presigned URL" >&2
+                    echo "[entrypoint] storage response body (first 2KB):" >&2
+                    head -c 2048 "$_out" 2>/dev/null >&2 || true
+                    echo >&2
+                    rm -f "$_out"
+                    return 1
+                    ;;
+            esac
             ;;
         200)
             # No redirect — re-fetch with output (rare, but handle it)
-            curl -sSf -o "$_out" "$@"
+            if ! curl -sS -o "$_out" "$@"; then
+                echo "[entrypoint] download: direct fetch failed (HTTP $_code)" >&2
+                return 1
+            fi
             ;;
         *)
-            echo "[entrypoint] Unexpected HTTP $_code" >&2
+            echo "[entrypoint] download: unexpected HTTP $_code from ${*##* }" >&2
+            echo "$_headers" | head -1 >&2
             return 1
             ;;
     esac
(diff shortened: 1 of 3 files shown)


mattrobinsonsre:
Strengthened test coverage: added an **always-on offline** guard (`test_presigned_url_is_sigv4_offline`) — `generate_presigned_url` signs locally, so with static fake creds it asserts the real signing path emits `X-Amz-Algorithm=AWS4-HMAC-SHA256` (never `AWSAccessKeyId=`) with **no AWS and no LocalStack**, so it runs in plain CI. Plus a discriminating negative (a client forced to the legacy `s3` signer *does* emit `AWSAccessKeyId=`) proving the assertion isn't vacuous. The AWS-side SSE-KMS rejection itself isn't offline-reproducible (LocalStack/moto don't enforce SigV2-vs-KMS); guaranteeing we never emit SigV2 is the complete substitute. **1490 passed.**

Issue #350, opened 2026-05-20
Fleet hooks for cross-workspace state consumers (bulk-update + autodiscovery templates) — follow-up to #344


mattrobinsonsre:
## Summary

Wire the producer-controlled consumer allowlist shipped in #344 into the two existing fleet-ergonomic surfaces: the **bulk-update endpoint** (#318) and the **autodiscovery rule template** (#318/#283). Same security invariant applies — admin/write on the producer is required, enforced server-side regardless of the mutation path.

This is the "define once → apply across the fleet" hook. The core feature is already operable via the API and the standalone provider resource (both shipped in #344); this issue makes large producer-consumer fan-out tractable to declare at scale.

## Background

#344 ships:
- `workspace_remote_state_consumers` junction table + producer-controlled CRUD endpoints.
- Runner-token authz branch on the v2 state endpoints (gates `terraform_remote_state` in agent mode).
- `terrapod_remote_state_consumer` provider resource (standalone, single edge).

Real-world adoption typically involves a small number of producers (shared modules, foundation networking, etc.) and many consumers (per-env workspaces). Declaring those edges one-by-one — via the API, the standalone provider resource, or the UI panel (#349) — works but doesn't scale gracefully. The two ergonomic surfaces this issue extends:

- **Bulk-update (#318)** — server-side workspace search + atomic settings update; the right hook for "set X across this filtered set of workspaces."
- **Autodiscovery rule templates (#283/#318)** — per-rule template fields applied to every workspace materialised from that rule; the right hook for "every new workspace this rule creates may consume X."

## Scope

1. **Bulk-update body**: accept a `remote_state_consumers` field in the per-workspace settings update. Semantics = declarative PUT replace-set (same as the existing endpoint), atomic within the bulk transaction. Per-workspace **producer-admin** check applies even within the bulk operation — a caller who is admin on workspace A but not on workspace B cannot set B's consumers via bulk-update; the whole bulk transaction fails (consistent with #318's all-or-nothing semantics). Honour dry-run by default.

2. **Autodiscovery rule template field**: add `remote_state_consumer_templates` (set of workspace IDs / selector) to the rule. On materialisation of a new workspace from this rule, the rule's consumer list is applied to the new workspace's allowlist as if the rule's owner had set them. The rule owner must hold producer-admin on every workspace referenced (validated at rule create/update, not at materialisation — fail fast).

[...]

Issue #349, opened 2026-05-20
UX: Remote State Sharing panel on workspace detail (follow-up to #344)


mattrobinsonsre:
## Summary

Add a "Remote State Sharing" panel to the workspace detail page (`web/src/app/workspaces/[id]/page.tsx`) consuming the cross-workspace consumer-allowlist API shipped in #344.

## Background

#344 ships the producer-controlled consumer allowlist for `terraform_remote_state` end-to-end at the API + provider + tests + docs layers. The UX panel for it was scoped out of that PR to avoid mixing a substantial UI integration (~200 lines on a 2935-line page including state hooks, SSE wiring, permission gating, and autocomplete) with a security-bearing backend PR. The standalone `terrapod_remote_state_consumer` provider resource covers the operator path in the meantime.

## Scope

A new section (or a small new tab) on the workspace detail page with two views:

- **Outbound (producer view):** workspaces I share my state to. Add by name (autocomplete from existing workspaces), remove. Add/remove are gated on producer admin/write permission; the panel is read-only otherwise (no buttons rendered). Calls:
  - `GET /api/terrapod/v1/workspaces/{id}/remote-state-consumers?filter[remote-state-consumer][type]=outbound`
  - `POST /api/terrapod/v1/workspaces/{id}/remote-state-consumers`
  - `DELETE /api/terrapod/v1/remote-state-consumers/{rsc_id}`
- **Inbound (consumer view, read-only):** workspaces whose state I'm authorized to read. Useful for traceability. Calls `GET … ?filter[type]=inbound`.

## Design notes

- Tab vs section: pick whichever fits the existing page rhythm — likely a section appended to the State tab (state-thematic) rather than a brand-new top-level tab, to keep the tab bar tight.
- Autocomplete: reuse whatever pattern the existing label / identity autocompletes use.
- Empty state: "This workspace's state is not shared with any other workspace." for outbound; clearer wording on inbound ("This workspace doesn't read state from any other workspace via terraform_remote_state.").
- Surface the `created-by` and `created-at` attributes from the API (already returned) for traceability.
- Error handling: 403 on mutate → "Only the producer's admin may grant or revoke remote-state access" message; do not show the buttons in the first place if perm < admin/write, but be robust to a stale permission cache.

## Tests

- Playwright: a producer admin sees the panel with Add/Remove controls; a read-only viewer sees the list but no controls. Add → row appears. Remove → row gone. Inbound view renders.

## Non-goals

[...]

Issue #348, opened 2026-05-20
Provider: terrapod_workspace.remote_state_consumers set attribute (follow-up to #344)


mattrobinsonsre:
## Summary

Add a `remote_state_consumers` set attribute to the existing `terrapod_workspace` provider resource so a single-config GitOps setup can declare the producer's consumer allowlist (#344) co-located with the producer's resource block.

The standalone `terrapod_remote_state_consumer` resource (shipped with #344) already covers the cross-config / cross-team case, where producer and consumer live in different Terraform configurations. This issue is the ergonomic complement for the *single-config* case: the user declares `terrapod_workspace.producer` and lists its consumers in one place.

Server-side, both shapes go through the same producer-admin authz check (locked in #344 — a consumer team cannot self-grant via either shape).

## Design

- New attribute on `terrapod_workspace`: `remote_state_consumers = set(string)` — workspace IDs of authorized consumers; Optional + Computed.
- Schema documented with an explicit caveat: **do not mix this attribute with standalone `terrapod_remote_state_consumer` resources targeting the same producer** — the workspace resource will see drift on every plan and fight the standalone resource.
- Reads via `GET /api/terrapod/v1/workspaces/{id}/remote-state-consumers?filter[remote-state-consumer][type]=outbound`.
- Writes via `PUT /api/terrapod/v1/workspaces/{id}/remote-state-consumers` (the declarative replace-set endpoint already in place).

## Subtleties to get right

This is the part that earned the follow-up rather than being folded into #344's bundle:

- **Null vs empty distinction.** `null` (attribute absent from config) must mean "I'm not managing this" — Read populates state from server, plan doesn't drift. `[]` (explicitly empty) must mean "remove all" — PUT an empty set.
- **Optional + Computed semantics.** When config is `null`, terraform uses the state value during plan, so no spurious diff against a server side managed by something else.
- **Drift when mixed with standalone resource.** Acceptable but must be documented — anyone mixing the two will see the workspace resource churn the set on every apply.
- **Plan-time validation.** Reject the case where the set contains the producer's own ID (self-reference is meaningless — the workspace already reads its own state).

## Tests (provider acceptance)

- Apply with `remote_state_consumers = [ws_a, ws_b]` → both grants present on server.
- Change to `[ws_a]` → ws_b grant removed; ws_a remains.
- Change to `[]` → all grants removed.
- Remove the attribute from config (terraform sees null) → no further mutation; existing grants left intact.
[...]

Issue #346, opened 2026-05-20
Replace runner backend-strip awk with override.tf (override files, not regex)


mattrobinsonsre:
## Summary

Replace the runner entrypoint's awk-based `cloud {}` / `backend "x" {}` block-stripping (`docker/runner-entrypoint.sh`) with **terraform-native override files**. The current approach edits user code in place using an anchored regex with depth tracking; that's structurally fragile and there's a documented native mechanism for exactly this purpose.

## Why the current strip is fragile

The awk pattern works for the common shapes but breaks (or risks breaking) on:

- `backend "remote"` followed by `{` on the **next line** (block opener on a separate line).
- Comments on the opener line (`cloud { # opt`) or before the `{`.
- String literals or heredocs inside the block that contain `{` / `}` — the depth tracker miscounts.
- Any future HCL construct that puts `cloud` / `backend` at column 0 in a position the regex didn't anticipate.

It also modifies the user's code in place before terraform/tofu sees it, which makes "why does my plan differ from local" hard to diagnose.

## Design

Drop the awk; in the working directory, write a **uniquely-named override file**:

```
_terrapod_backend_override.tf
```

Content:

```hcl
# Terrapod runner: force local backend for in-runner execution.
# Override files (*_override.tf) are merged by terraform/tofu with
# replacement semantics over the main config.
terraform {
  backend "local" {}
}
```

Two non-obvious choices, both deliberate:

1. **Unique filename, leading underscore.** Terraform recognises `*_override.tf` as an override file (matched by suffix), so this name *is* an override. The leading-underscore + project-prefix makes a clash with a user-authored file vanishingly unlikely, so writing it **never overwrites a user's existing `override.tf`** or any other file. This is the one thing the implementation must not get wrong.
2. **Detect-and-defer if the user already has their own backend-neutralising override.** Before writing ours, scan the workdir's `override.tf` and `*_override.tf` files for a `terraform { backend ... }` or `terraform { cloud {} }` block that already neutralises remote execution. If one is present, the user has done the work — leave them alone, skip writing ours, and log the decision so it's visible in the runner log. This trusts user code and avoids a two-override conflict on the same block.

If neither condition is met (no pre-existing user override of the backend), write ours.

## Spike (required before implementing)

A small spike on the live Tilt stack against **both terraform and tofu** with these fixture configs in the runner workdir, observed end-to-end:

1. Main config with `terraform { backend "remote" {} }` — our override should displace it.
[...]

PR #351, opened 2026-05-20
feat: cross-workspace terraform_remote_state — producer-controlled consumer allowlist (closes #344)

### LINKED ISSUE #344: Cross-workspace terraform_remote_state — producer-controlled consumer allowlist
## Summary

Make `data "terraform_remote_state"` work for **cross-workspace** state reads against Terrapod (one workspace consuming another's state/outputs). The read plumbing already exists; the work is a **producer-controlled explicit consumer allowlist** plus the agent-mode authorization wiring, verification, and docs.

## What already works

`terraform_remote_state` with the `cloud`/`remote` backend uses CLI-contract endpoints Terrapod already serves (`docs/tfe-cli-surface.md`):

- `GET /api/v2/organizations/default/workspaces/{name}` → id
- `GET /api/v2/workspaces/{id}/current-state-version` → `hosted-state-download-url`
- `GET /api/v2/state-versions/{id}/download` → raw state JSON; terraform extracts `outputs` client-side

**Local execution mode already works today** for a user whose API token has `plan` on the producer workspace. `terraform_remote_state` downloads the full state and extracts outputs itself — no outputs store is needed (the outputs-only path is `tfe_outputs`, which needs the excluded `tfe` provider).

## The gap (agent mode) + the chosen model

`current-state-version` requires `read`, `download` requires `plan` on the **producer** (correct — raw state holds secrets). In agent mode the consumer run's Job uses a runner token scoped to its own `run_id` with only the `everyone` role, so cross-workspace reads 403. Effectively blocked in agent mode.

**Design decision (locked): producer-controlled explicit consumer allowlist.** Each workspace owns an explicit list of other workspaces permitted to read its state. Default empty (not shared) — secure by default. Control stays with the state owner.

Rejected alternatives (record in the issue, do not revisit):
- **Global / org-wide share flag** (TFE's `global-remote-state`) — too crude for a single-org self-hosted deployment ("every workspace can read your secrets").
- **Consumer-side authorization** (owner-derived perms, or role-assigned-to-workspace) — owner coupling + transitive-grant surprise; control must stay with the producer, not the consumer or an admin grant.

This matches TFE's "share with specific workspaces" (`remote-state-consumers`) and intentionally omits TFE's `global-remote-state`.

## Proposed design (Terrapod-native)

- **DB**: junction table `workspace_remote_state_consumers (producer_workspace_id FK, consumer_workspace_id FK, unique pair)` — cascades on workspace delete; queryable both directions. No rows ⇒ not shared.
- **Authorization split**:
  - User / API-token principals (local CLI, UI, automation): **unchanged** — existing per-user label-RBAC (`plan` on producer).
  - Runner-token principals (agent runs): grant `current-state-version` + `download` for producer A **iff** the token's run → consumer workspace B is in A's consumer allowlist. Only new check; producer-controlled.
- **Management API**: Terrapod-native (`/api/terrapod/v1/...`, NOT `tfe_v2.py` — this is management, not CLI-consumed). Add/remove consumers by workspace (accept name, store id — single-org). Audited as a workspace setting change.
- **UI**: workspace settings panel listing/editing consumers.
- **Fleet scale**: settable via the bulk-update endpoint (#318) and autodiscovery rule templates, so a monorepo rule can declare "workspaces I create may consume `shared-network`" without per-workspace edits.
- **Provider (Infrastructure-as-Code) — primary at-scale management**: the Terrapod provider is the main answer for cross-talk-heavy orgs that define workspaces as code. Ship two shapes, both gated by the same server-side check below:
  - `remote_state_consumers` attribute on the existing `terrapod_workspace` resource (producer-centric; edges co-located with the producer's block; common GitOps case).
  - a standalone `terrapod_remote_state_consumer` resource (`producer`/`consumer`) for setups where producer and consumers live in separate configs/teams.

### Security invariant (non-negotiable)

Creating or removing a consumer edge **requires admin/write on the PRODUCER workspace**, enforced in the API, regardless of which Terraform config, module, or UI initiates it. This is what keeps control with the state owner and makes the provider shape a pure ergonomics choice (a consumer team applying a standalone edge resource gets 403 unless they hold producer-admin — they must route it through the producer owner).
- Single-org; label-RBAC unchanged; distributed-scheduler-safe (no new background work).

## Acceptance criteria

- Local mode: documented + verified end-to-end (consumer reads producer outputs with a `plan`-permissioned token).
- Agent mode: with the producer listing the consumer, the consumer run's runner token can read `current-state-version` + `download`; not listed ⇒ 403 (secure by default).
- Allowlist CRUD is audited; cascades on workspace delete; mutation requires producer-workspace admin/write (enforced server-side, provider/UI/bulk alike); settable via bulk-update + autodiscovery templates.
- Provider parity: `terrapod_workspace.remote_state_consumers` attribute + standalone `terrapod_remote_state_consumer` resource, both rejected without producer-admin; provider docs + acceptance tests.
- Tests (authz matrix: user/runner × listed/not-listed × local/agent; cascade) + docs (`api-reference.md`, a remote-state composition guide) + runbook ("cross-workspace state read 403 / how to authorize a consumer").

## Relationship to run triggers (decided: independent)

Run triggers (existing: `RunTrigger`, `routers/run_triggers.py`) and the remote-state consumer allowlist are **independent sibling composition edges; neither implies the other.**

- Run trigger = consumer-owned *"re-run me when A applies"* (ordering/causality).
- Remote-state grant = producer-owned *"B may read my state"* (data + authorization).

They have correctly-opposite ownership (each knob owned by the party that should own it) and may share table/listing/provider/UX *patterns*, but stay separate with independent authz. Setting a trigger never auto-creates a state grant.

**Canonical justification (docs example):** workspace A manages a VPC; workspace B reads it via `data "aws_vpc"` from the cloud provider (not `terraform_remote_state`). B wants a run trigger A→B (re-plan after A changes infra) but must NOT get a state grant — it never reads A's state, and granting it would needlessly expose A's secrets. Coupling the two would over-permission the common data-source dependency.

## Non-goals

- `tfe_outputs` provider data source (excluded `tfe` provider).
- Org-wide/global sharing; project/team scoping; consumer-side or owner-derived authorization.

---

### Phase 0 — Spike / de-risk ✅ COMPLETE

- **Own-state path is separate (regression risk eliminated):** agent runs read/write their own state via `/api/terrapod/v1/runs/{id}/artifacts/state` (`run_artifacts.py`, run-scoped runner auth) — a different code path from the v2 `current-state-version`/`download` endpoints. Adding cross-workspace authz to the v2 endpoints cannot regress normal agent runs; the v2 state endpoints are exclusively the `terraform_remote_state`/go-tfe path.
- **Runner-token principal ready:** `get_current_user` already sets `auth_method="runner_token"` + `run_id` (`dependencies.py`). Authz branch is direct: runner token → `run_id`→`Run.workspace_id` (consumer) → allow iff `(producer,consumer) ∈ junction`.
- **Runner credentials already cover the in-scope case:** entrypoint writes `credentials "$MIRROR_HOST" {…}` (Terrapod API host) with the runner token; same-instance `terraform_remote_state` backend host == `MIRROR_HOST`, so likely no entrypoint change (verify in smoke). Cross-instance is out of scope.
- **Run-trigger reuse: mirror, don't share.** `run_triggers.py`/`RunTrigger` give the exact edge patterns; `RunTrigger`'s docstring already states "No data is passed — downstream read outputs via terraform_remote_state independently." Consumer-allowlist router structurally parallels it; sole difference = producer-side authz.
- **NEW RISK (most likely silent breakage):** the entrypoint strips `cloud {}` / `backend "…" {}` blocks from uploaded config. A `data "terraform_remote_state"` block uses `backend = "remote"` as an *argument*, not a `backend {}` block — the stripping logic must be proven not to mangle it. → Phase 2 guard + Phase 8 smoke assertion.

### Phase 1 — Database
- `workspace_remote_state_consumers` table (`producer_workspace_id` FK→workspaces ON DELETE CASCADE, `consumer_workspace_id` FK ON DELETE CASCADE, `created_at`, `created_by`; UNIQUE(producer,consumer); index both FKs). Model + Workspace relationship helpers. Alembic migration (hash revision; `down_revision` = current head; server_default+python default conventions). Non-breaking (empty ⇒ no behaviour change).

### Phase 2 — API
- Terrapod-native router mirroring `run_triggers.py`: `GET` list (inbound/outbound), `PUT` replace-set, `POST`/`DELETE` single edge. Accept consumer by name→id (single-org). Reject self-ref (422); idempotent.
- **Security invariant:** all mutations require admin/write on the **producer** (`resolve_workspace_permission`); list requires read on producer.
- Read-path authz: in `current_state_version`/`download_state` (tfe_v2.py) add a runner-token branch — resolve run→consumer; allow iff `(producer,consumer) ∈ junction`. Existing user/API-token RBAC path untouched.
- **Config-stripping guard:** prove/adjust the entrypoint block-strip so `data "terraform_remote_state"` survives intact (regex targets `cloud {`/`backend "x" {` blocks, not the `backend =` argument); add a unit/golden test on the strip logic.
- **Audit the read** (cross-workspace state consumption), not just the grant; audit allowlist mutations.

### Phase 3 — Provider
- `terrapod_workspace.remote_state_consumers` set attribute (→ PUT replace; order-independent diff) + standalone `terrapod_remote_state_consumer` resource (→ POST/DELETE). Both rejected server-side without producer-admin. Docs/examples + Go acceptance tests; `gofmt`/`vet`/`build` verified manually.

### Phase 4 — UX
- Workspace settings "Remote State Sharing" panel (list/add-by-name/remove; gated on producer admin/write; 403 handling); optional read-only "consumes from…" reverse view. Presented alongside run triggers as a combined "Workspace dependencies" surface (independent edges, surfaced together). Types + `npm run build` verified.

### Phase 5 — Fleet hooks
- Wire consumer-list into bulk-update (#318) + autodiscovery rule templates, reusing the shared spec-validation pattern (like `run_task_templates`). Provider remains primary at-scale path.

### Phase 6 — Docs / CLAUDE.md / schema
- `api-reference.md` (mgmt endpoints + cross-workspace read rule); new `docs/remote-state.md` (local vs agent; **security: full state incl. secrets exposed**; the data-source-on-managed-infra example justifying trigger≠grant); `runbooks.md` (403 → authorize a consumer; producer-delete blast radius); `autodiscovery.md`/bulk docs (template hook); `index.md` feature row; `tfe-cli-surface.md` note (no new v2 routes — authz logic only); local CLAUDE.md (state arch + API↔UX map + routing; gitignored, not committed). Confirm **no Helm values change** (default-off via empty list).

### Phase 7 — Tests
- Authz matrix: principal (user/API/runner) × listed/not-listed × local/agent × **own-state regression** × producer-admin-for-mutation; junction cascade (delete consumer/producer); name→id; idempotent PUT; self-ref reject; **config-strip golden test (terraform_remote_state survives)**. API + provider acceptance tests. `make test` + lint green.

### Phase 8 — Smoke (live, adversarial — mandatory gate)
- Happy local; happy agent (producer lists consumer → runner reads → plan shows consumed value via Job log + plan output).
- Negative/security: consumer not listed → runner 403, **zero state bytes**; X reading Y where X∉Y.consumers → 403; **regression: normal agent run reads own state**; non-producer-admin self-grant attempt (user/runner/consumer-team, incl. provider standalone resource) → 403; cascade on delete; **config-strip: a real `data "terraform_remote_state"` block survives the entrypoint and resolves**.
- Honest coverage statement in the PR.

### Cross-cutting / not-yet-considered
- **Listeners: no change** (stateless; don't touch state) — stated explicitly.
- **Runners:** likely no change (Phase 0); the only runner-side risk is the config-strip (Phase 2 guard).
- Full-state (not outputs-only) exposure documented; outputs-only = future hardening, not this issue.
- No lock on cross-reads (read-only) — confirm + doc. Stale/empty producer state, cycles (A↔B), archived-producer-still-consumable-until-state-purged (ties to #314) — document expectations.
- Runner-token TTL vs large `init` resolving many remote_state sources (1h/2h) — note. Scope = `remote`/`cloud` backend data source only (not `http`).
- Backwards compatible / non-breaking; **no release until Phase 8 passes**; bundled single PR; minor bump with pre-release audit D+F.

### Sequencing
Phase 0 ✅ → 1 → 2 (+7 alongside) → 3/4/5 (parallel) → 6 → 8 (gate) → bundled PR → merge when green (no release) → fold into next minor with audit.

mattrobinsonsre:
Closes #344. Phase 0 spike + Phases 1, 2, 3 (provider), 6, and 7 of the implementation plan in the issue. Phases 4 (UX), 5 (fleet hooks), and the workspace-resource set attribute are deliberately split into focused follow-ups (#349, #350, #348) to keep this security-bearing PR reviewable; the core feature is independently operable via the API + standalone provider resource.

## What ships

- **Producer-controlled consumer allowlist.** Junction table `workspace_remote_state_consumers` (FK cascade both sides); default empty ⇒ not shared (secure by default).
- **Terrapod-native management endpoints** under `/api/terrapod/v1/workspaces/{id}/remote-state-consumers` (list inbound/outbound, POST single, PUT declarative replace-set, DELETE single by id). All mutations require **admin/write on the PRODUCER** — a consumer team cannot self-grant.
- **Runner-token authz branch** on the existing CLI-contract endpoints `GET /api/v2/workspaces/{id}/current-state-version` and `GET /api/v2/state-versions/{id}/download`. Granted iff the runner's own workspace is the producer (self-read; harmless) or is in the producer's allowlist. User / API-token RBAC paths are completely unchanged.
- **Terrapod provider standalone resource** `terrapod_remote_state_consumer` (model + resource; immutable: create + read + delete). Same producer-admin authz applies via the same API.
- **Documentation**: new `docs/remote-state.md` composition guide (incl. the data-source-on-managed-infra example justifying why run triggers and remote-state grants are independent), `api-reference.md` updates, runbook entry for the 403 case, `index.md` feature row + doc link.

## Key design decisions (locked, recorded in the issue)

- **Producer-controlled allowlist** — explicit per-edge list owned by the state owner. No `global-remote-state`-style toggle (too crude for single-org). No consumer-side / owner-derived / workspace-role authorization (control must stay with the producer; secret-exposure axis is separate from RBAC).
- **Independent from run triggers.** Neither edge implies the other. A workspace can need a run trigger without a remote-state grant (consuming managed infra via `data "aws_*"`), and vice versa. Documented as the canonical example.
- **Same security invariant across provider, UI, bulk, and direct API**: server-side `admin/write` on the producer is the single gate.

## Phase 0 findings (de-risking; recorded in the issue)

- **Own-state regression risk eliminated**: agent runs read their own state via `/api/terrapod/v1/runs/{id}/artifacts/state` (run-scoped runner auth) — a different code path from the v2 endpoints this PR modifies. The v2 endpoints are exclusively the `terraform_remote_state`/go-tfe path.
- **Mirror, don't share** the run-trigger machinery: `RunTrigger`'s own docstring already states *"No data is passed — downstream read outputs via terraform_remote_state independently"*. The consumer router structurally parallels run-triggers.
- **Config-stripping risk**: the entrypoint's `cloud {}` / `backend {}` strip pattern targets line-leading `cloud`/`backend` openers with a `{`. A `data "terraform_remote_state"` block starts with `data` and uses `backend = "x"` as an argument (no `{`), so the strip does not mangle it. **Regression-guarded by a golden test** that runs the real entrypoint awk against fixtures (`tests/runner/test_config_strip.py`). The strip's broader fragility is its own follow-up: #346.

## Tests

- **1509 pass**, lint clean.
- 7 unit cases on the security-critical `_runner_state_read_allowed` helper (non-runner principals fall through; missing/invalid run_id denied; self-read granted; listed-consumer granted; **not-listed-consumer denied — the core security gate**).
- 6 HTTP CRUD cases on the management router (admin happy path; non-admin 403 on mutations; self-reference 422; duplicate 409; invalid filter 422; 404 on delete).
- 3 strip golden tests (removes cloud/backend; **preserves `data "terraform_remote_state"`**; mixed file).

## Honesty / coverage gap

The smoke section below covers the live end-to-end run. UX panel (#349), workspace-resource set attribute (#348), and fleet hooks (bulk-update + autodiscovery templates, #350) are explicitly scoped out and tracked.

## Smoke (live Tilt — gated; in progress)

The adversarial smoke from the plan runs against the live stack with the rebuilt API + runner image. Will append results to this PR before merge:

- Happy local mode: producer w/ outputs → consumer `terraform_remote_state` w/ plan-permissioned user token → outputs read.
- Happy agent mode: producer lists consumer → consumer runner reads producer state → plan resolves consumed value.
- **Negative (the core)**: consumer NOT listed → runner 403, **zero state bytes leak**.
- Regression: normal agent run still reads its own state via `run_artifacts` (must-not-break).
- Non-producer-admin self-grant attempt via API and provider → 403.
- Cascade: delete consumer → edge gone; delete producer → edges gone.

## Follow-ups filed

- #348 `terrapod_workspace.remote_state_consumers` set attribute (provider, single-config GitOps complement)
- #349 UX panel on workspace detail
- #350 Fleet hooks (bulk-update + autodiscovery rule template)
- #346 Replace runner backend-strip with `override.tf` (related fragility surfaced during Phase 0)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,15 @@ jobs:
         run: poetry export -f requirements.txt -o requirements.txt --without-hashes
         working-directory: services
       - name: Run pip-audit
-        run: pip install pip-audit && pip-audit -r services/requirements.txt --strict --desc
+        # PYSEC-2025-183 (pyjwt): disputed by the supplier — the advisory
+        # concerns weak HMAC keys, but key length is controlled by the
+        # application that uses the library, not by the library. No fix
+        # version is published. Suppress until a fix lands or the dispute
+        # is resolved upstream.
+        run: |
+          pip install pip-audit
+          pip-audit -r services/requirements.txt --strict --desc \
+            --ignore-vuln PYSEC-2025-183
 
   # ── Dependency Audit (npm) ────────────────────────────
   npm-audit:
--- a/alembic/versions/42266b856e8e_workspace_remote_state_consumers.py
+++ b/alembic/versions/42266b856e8e_workspace_remote_state_consumers.py
@@ -1,0 +1,68 @@
+"""workspace remote state consumers (#344)
+
+Producer-controlled allowlist authorizing a consumer workspace's agent
+runs to read a producer workspace's state via terraform_remote_state.
+Empty for a producer => its state is not shared (secure by default);
+no behaviour change for existing deployments.
+
+Revision ID: 42266b856e8e
+Revises: 3f0ad398be2c
+Create Date: 2026-05-20
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "42266b856e8e"
+down_revision: str | None = "3f0ad398be2c"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "workspace_remote_state_consumers",
+        sa.Column("id", sa.UUID(as_uuid=True), primary_key=True),
+        sa.Column(
+            "producer_workspace_id",
+            sa.UUID(as_uuid=True),
+            sa.ForeignKey("workspaces.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "consumer_workspace_id",
+            sa.UUID(as_uuid=True),
+            sa.ForeignKey("workspaces.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.func.now(),
+            nullable=False,
+        ),
+        sa.Column("created_by", sa.String(255), server_default="", nullable=False),
+        sa.UniqueConstraint(
+            "producer_workspace_id",
+            "consumer_workspace_id",
+            name="uq_workspace_remote_state_consumers",
+        ),
+    )
+    op.create_index(
+        "ix_wrsc_consumer_workspace_id",
+        "workspace_remote_state_consumers",
+        ["consumer_workspace_id"],
+    )
+    op.create_index(
+        "ix_wrsc_producer_workspace_id",
+        "workspace_remote_state_consumers",
+        ["producer_workspace_id"],
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("ix_wrsc_producer_workspace_id", table_name="workspace_remote_state_consumers")
+    op.drop_index("ix_wrsc_consumer_workspace_id", table_name="workspace_remote_state_consumers")
+    op.drop_table("workspace_remote_state_consumers")
--- a/docker/Dockerfile.test
+++ b/docker/Dockerfile.test
@@ -28,6 +28,11 @@ RUN poetry config virtualenvs.create false \
 COPY services/terrapod ./terrapod
 COPY services/tests ./tests
 
+# Copy the runner entrypoint so tests can exercise its real shell logic
+# as a fixture (e.g. the cloud/backend block-stripping golden test for
+# #344). The entrypoint itself isn't executed here — it's read.
+COPY docker/runner-entrypoint.sh ./docker/runner-entrypoint.sh
+
 # Storage data directory for tests
 RUN mkdir -p /var/lib/terrapod/storage
 
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -386,6 +386,8 @@ GET /api/v2/workspaces/{id}/current-state-version
 
 **Required permission:** `read` on the workspace.
 
+For **agent-mode runs** reading another workspace's state via `data "terraform_remote_state"` (runner-token principals), authorization is by the producer-controlled consumer allowlist instead of per-user RBAC — see [Cross-Workspace Remote-State Consumers](#cross-workspace-remote-state-consumers) and [the composition guide](remote-state.md).
+
 ### Create State Version
 
 ```
@@ -423,6 +425,8 @@ GET /api/v2/state-versions/{id}/download
 Returns a redirect to a presigned URL for the raw state file.
 
 **Required permission:** `plan` on the workspace.
+
+For **agent-mode runs** reading another workspace's state via `data "terraform_remote_state"` (runner-token principals), authorization is by the producer-controlled consumer allowlist instead of per-user RBAC — see [Cross-Workspace Remote-State Consumers](#cross-workspace-remote-state-consumers).
 
 ### Upload State Content
 
@@ -668,6 +672,83 @@ GET /api/terrapod/v1/runs/{run_id}/apply
 ```
 
 Returns apply metadata and log download URL.
+
+---
+
+## Cross-Workspace Remote-State Consumers
+
+Producer-controlled allowlist of workspaces authorized to read this workspace's state via `data "terraform_remote_state"`. Default is empty (not shared) — secure by default. All mutations require **admin/write on the producer** (the state owner). Independent of run triggers — see [the composition guide](remote-state.md).
+
+When a runner-token principal (agent-mode run) hits `/api/v2/workspaces/{id}/current-state-version` or `/api/v2/state-versions/{id}/download` for another workspace, authorization is by this allowlist instead of per-user RBAC. User / API-token principals (CLI, UI, automation) continue through the existing per-user RBAC path.
+
+### List Consumers
+
+```
+GET /api/terrapod/v1/workspaces/{id}/remote-state-consumers?filter[remote-state-consumer][type]=outbound
+```
+
+`outbound` (default) — workspaces this workspace shares its state to (this workspace is the producer).
+`inbound` — workspaces whose state this workspace is authorized to read (this workspace is the consumer).
+
+**Required permission:** `read` on the workspace.
+
+### Authorize a Consumer
+
+```
+POST /api/terrapod/v1/workspaces/{producer_id}/remote-state-consumers
+```
+
+**Request body:**
+```json
+{
+  "data": {
+    "relationships": {
+      "consumer": {"data": {"id": "ws-CONSUMER", "type": "workspaces"}}
+    }
+  }
+}
+```
+
+**Required permission:** `admin` on the **producer** workspace. A consumer team cannot self-grant.
+
+Errors: `422` self-reference; `409` already authorized; `422` over the per-producer cap.
+
+### Replace Consumer Set (Declarative)
+
+```
+PUT /api/terrapod/v1/workspaces/{producer_id}/remote-state-consumers
+```
+
+Idempotent declarative replace of the producer's full consumer set in one atomic transaction. Supports the Terrapod provider's set-valued attribute. Body: `{"data": [{"type": "workspaces", "id": "ws-..."}, ...]}`.
+
+**Required permission:** `admin` on the producer.
+
+### Show Consumer Grant
+
+```
+GET /api/terrapod/v1/remote-state-consumers/{id}
+```
+
+**Required permission:** `read` on the producer.
+
+### Revoke a Consumer Grant
+
+```
+DELETE /api/terrapod/v1/remote-state-consumers/{id}
+```
+
+**Required permission:** `admin` on the producer. A consumer cannot self-revoke.
+
+### Consumer Grant Response Attributes
+
+| Attribute | Description |
+|---|---|
+| `producer-workspace-name` | Name of the producer workspace |
+| `consumer-workspace-name` | Name of the consumer workspace |
+| `created-at` | RFC3339 timestamp the grant was created |
+| `created-by` | Identity that created the grant |
+
+Plus relationships `producer` and `consumer` (both `workspaces`-typed).
 
 ---
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -34,6 +34,7 @@ Terrapod is **not** a fork of Terraform or OpenTofu. It orchestrates them.
 | **Binary Caching** | Pull-through cache for terraform/tofu CLI binaries |
 | **Workspace Autodiscovery** | Atlantis-style monorepo autodiscovery with rule templating; safe-by-default rename/delete/orphan lifecycle (opt-in destroy) |
 | **Bulk Workspace Operations** | Server-side workspace search + all-or-nothing bulk settings update (dry-run by default; never triggers runs) |
+| **Cross-Workspace Remote State** | `terraform_remote_state` composition with a producer-controlled consumer allowlist (secure by default; secret-bearing state stays with its owner) |
 
 ---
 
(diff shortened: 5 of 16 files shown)


mattrobinsonsre:
## Phase 8 — live adversarial smoke ✅

Ran against the live Tilt stack (`terrapod-api-585f847b64-jjrv6`, alembic head `42266b856e8e`, migration applied; junction table created empty). Producer = `smoke-306` (workspace with state), Consumer = `smoke-cli-tofu-v024`. Runner tokens minted via `auth.runner_tokens.generate_runner_token` scoped to a real run in each consumer workspace.

| # | Scenario | Result |
|---|---|---|
| 1 | Admin POST grant authorizing consumer → 201 with expected JSON:API body | ✓ |
| 2 | Runner token (consumer's run) reads producer's `current-state-version` with grant in place | **HTTP 200** ✓ |
| 3 | Admin DELETE grant → 204; re-read with same runner token | **HTTP 403**, body 50 bytes (error JSON — **zero state bytes leak**) ✓ |
| 4 | Cross-workspace isolation: runner from `gamma-network` (not in producer's allowlist) reads producer | **HTTP 403** ✓ |
| 5 | Self-reference grant rejected at management endpoint | **HTTP 422** ✓ |
| 6 | Regression: runner reads its OWN workspace state via the v2 path (self-read shortcut works) | **HTTP 200** ✓ |

The core security invariant is proven end-to-end on the real wire: **a runner token can read another workspace's secret-bearing state only when the producer's admin has explicitly granted that consumer; revoking removes access immediately and leaks no bytes.**

## Coverage not run live (honest gap)

- **Cascade-on-workspace-delete**: enforced by FK `ON DELETE CASCADE` in the schema; covered by the design and the migration. Not live-tested because deleting real smoke workspaces is destructive.
- **Non-admin attempts to grant via API / provider**: covered by the HTTP CRUD unit test (`test_create_requires_producer_admin` → 403) and `test_delete_requires_producer_admin`. Not run live because that needs a second non-admin principal in the smoke env.
- **Full agent-mode `terraform_remote_state` plan against a real config**: the runner-token authz branch was exercised directly via curl with a real minted runner token (the *exact* same code path a runner would hit), so the security boundary is verified. A terraform-driven end-to-end is the UX of #349's domain rather than a security gate.

Ready to merge per the established pattern (CI green + smoke green). No release.

mattrobinsonsre:
## Response to third-party review (`055e82a`)

Substantive items addressed:

- **Should-fix #2 — audit the cross-workspace read** (the issue's Phase 2 explicitly contracted this): the runner-authz helper now emits an explicit `workspace.remote_state_read` audit row carrying the producer / consumer / grant / run quadruple when the grant fires. Middleware still records the HTTP path; the helper's row records the cross-workspace context the URL can't express. Self-reads stay unaudited (no cross-workspace context, would be noise).
- **Should-fix #5 — `test_create_happy_path` was passing against a no-op `db.refresh` mock**: `db.refresh` now has a `side_effect` that populates the row's `producer_workspace` / `consumer_workspace` the way the real DB session would, and the test asserts the producer/consumer names and ids in the response shape. Would now fail against a broken refresh.
- **Should-fix #7 — no HTTP-level test of the helper-to-endpoint wiring**: three new tests in `TestRunnerAuthzWiring` patch `_runner_state_read_allowed` and drive both `current_state_version` and `download_state` via the ASGI client, asserting the helper is awaited and the user-RBAC fallback is or isn't consulted depending on the helper's return. A future refactor that drops the helper call from either endpoint will fail these now.

Deliberately punted (with rationale, not silently dropped):

- **Should-fix #1 (historical SV downloads)**: matches the existing per-user `plan` semantics on `download` — a grant authorises read of the producer's full state history, not just the current SV. State version IDs are UUIDs (unguessable). Worth a deliberate line in `docs/remote-state.md` as the reviewer suggests — happy to take a tiny follow-up commit for that doc clarification if you want, but it's not a code change.
- **Should-fix #3 (PUT's N sequential workspace lookups)**: 200-cap means worst case is bounded; functional and atomic. Worth a one-shot `select(... .in_(desired))` optimisation but not security-critical and out of #344's scope; will matter more once #350 wires the consumer list through bulk-update.
- **Should-fix #4 (extract a `_state_read_authorized` helper)**: pure refactor; not changing behaviour. Skipping in this PR.
- **Should-fix #6 (`test_create_requires_producer_admin` uses "read" rather than "write" or `None`)**: read fails the `admin` check the same way write does — both branches of `has_permission(perm, "admin")` are False. Strengthening with `None` would be incrementally better but not catching a different defect. Skipping.

Nice-to-haves (8-13) intentionally deferred.

`make test`: **1512 pass**, lint clean. CI re-running on the new head.

PR #357, opened 2026-05-22
fix: use redirect-aware download for apply-phase lock file

the3venthoriz0n:
## Summary
- Apply-phase lock file download used raw `curl -o` which doesn't follow 302 redirects
- Artifact endpoints return 302 → presigned URL; without following, the file is empty
- Terraform ignores the empty lock file, resolves providers fresh, creating a lock file mismatch with the saved plan
- Switched to `tp_curl_download` (the same redirect-aware wrapper used by state/config/plan downloads)

## Test plan
- [x] Run agent-mode plan+apply with a provider dependency (`null_resource`)
- [x] Verify apply phase logs show "Reusing .terraform.lock.hcl from plan phase"
- [x] Verify apply phase init says "Reusing previous version of..." (not "Finding latest version...")
- [x] Verify apply completes successfully without "Inconsistent dependency lock file" error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docker/runner-entrypoint.sh
+++ b/docker/runner-entrypoint.sh
@@ -453,14 +453,12 @@ fi
 # older runner that didn't upload a lock file, or an older API without
 # the /lock-file endpoint.
 if [ "$TP_PHASE" = "apply" ] && [ -n "$TP_API_URL" ] && [ -n "$TP_RUN_ID" ]; then
-    LOCK_DL_HTTP=$(curl -sS -o .terraform.lock.hcl -w "%{http_code}" \
-        -H "$AUTH_HEADER" \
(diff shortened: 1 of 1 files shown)


mattrobinsonsre (approved):
Thanks @the3venthoriz0n — nice catch. This is a real bug that's been silently degrading apply-phase provider stability (the inconsistent-lock-file errors I'd bet you were chasing). The reason `tp_curl_download` exists is exactly the case this fixes: every other artifact in this entrypoint already goes through it, and the lock-file download was the one that wasn't — a regression from when the wrapper was introduced.

The diff is tight and uses the existing wrapper consistently (same `TP_LAST_HTTP` surface, same redirect-aware 302→presigned flow used by state/config/plan). Your test plan covers the visible symptom (the "Reusing previous version of..." vs "Finding latest version..." init line) which is the right thing to verify.

Merging now — we'll smoke-test on the live stack before the next release tag.

Issue #347, opened 2026-05-20
Terrapod migration tool — first sources TFE/HCP Terraform and Atlantis


mattrobinsonsre:
## Summary

Implement a migration tool that helps a customer move their Terraform platform layer **into Terrapod** from whichever competing platform they currently run. Two first sources, broad enough audience that this should be one tool with a pluggable source layer rather than N separate spikes:

- **TFE / HCP Terraform** — teams leaving because the bill no longer makes sense; the heavy case (rich platform state to import).
- **Atlantis** — teams wanting state, a UI, a registry, RBAC, and the rest that Atlantis doesn't provide; the light case (the source has almost no state of its own — config is in the repo, state files are already in the user's bucket, run logs live in PR comments).

A unified CLI keeps the operator UX consistent ("I'm moving onto Terrapod, here's my source, run the tool") and lets the framework-level decisions (dry-run model, idempotency, RBAC translation, distribution shape) be made once rather than re-litigated per platform.

Two phases inside this issue: **Phase 1** produces the design — the resource mapping tables per source, the shared intermediate-representation schema, and a one-workspace PoC per source so we know the design is implementable. **Phase 2** builds the production tool. Both phases close out under this issue; no separate follow-up.

## Architecture: pluggable sources, one writer

```
┌──────────────────────────┐
│ Source plugin            │   one per migration source
│ (--source=tfe|atlantis)  │   — TFE API client, atlantis.yaml parser, …
└─────────────┬────────────┘
              │ normalised intermediate representation
              ▼
┌──────────────────────────┐
│ Migration framework      │   dry-run, idempotency, mapping report,
│                          │   resumability, RBAC review gate
└─────────────┬────────────┘
              │
              ▼
┌──────────────────────────┐
│ Terrapod writer          │   single Terrapod API client + state push
└──────────────────────────┘
```

The intermediate representation is the Phase 1 design product — a plain-data description of "the things we found, by category" that any source can populate and the writer can consume. The framework knows nothing platform-specific; the source plugin knows nothing about Terrapod; the writer knows nothing about the source.

## Source 1 — TFE / HCP Terraform (one TFE org → the Terrapod single org)

Input: a TFE/HCP Terraform org name + an API token with read access to that org. Output: the Terrapod deployment populated with the corresponding resources. One org per run — TFE multi-org maps to multiple separate Terrapod deployments by Terrapod's design.

Resources to map, in priority order (high to low):

[...]

PR #359, opened 2026-05-22
feat: add external state mode for Atlantis migration coexistence

the3venthoriz0n:
## Problem

State migration is the single most painful part of adopting Terrapod. Today, to move a workspace from Atlantis to Terrapod you must:

1. Import all state files into terrapod (and hope nothing gets lost)
2. Change all `backend "s3" {}` blocks to `cloud {}` in every repo
3. Coordinate the cutover so nothing applies to the wrong state during transition
4. Disable Atlantis — with no easy way back if something goes wrong

This is a hard, irreversible cutover that blocks adoption for teams with hundreds of workspaces.

## Solution: External State Mode

This PR adds a workspace-level `state_mode` setting (`managed` | `external`) that lets Terrapod orchestrate runs while preserving the code's native backend block. The runner uses `backend "s3" {}` directly — same state file, same bucket, same DynamoDB lock as Atlantis.

**This enables:**
- **Side-by-side deployment** — run Terrapod and Atlantis against the same repos, switching between them per-workspace
- **Zero-migration onboarding** — point Terrapod at existing repos with no code changes, no state migration
- **Gradual adoption** — move workspaces one at a time, validate each one, roll back instantly if needed
- **Future bulk migration path** — once confident, bulk-import state into Terrapod and flip workspaces to managed mode

## Scope

- Currently supports S3+DynamoDB backends (the Atlantis standard)
- Other backends (GCS, Azure) can be added later with the same pattern
- This is phase 1 of a migration toolkit. Future work:
  - Bulk state import into Terrapod workspaces
  - Automated PR submission to swap backend blocks across repos

## How It Works

When `state_mode=external`:
- Runner preserves the user's `backend "s3" {}` block (no override file written)
- Skips state download from Terrapod and state upload after apply
- Terraform handles lock/read/write/unlock directly against S3+DynamoDB
- State endpoints return 404/409 (state not managed by Terrapod)
- Drift detection works naturally via `plan -detailed-exitcode`

When `state_mode=managed` (default):
- Unchanged — today's behavior

## Changes

- **Model:** `state_mode` on Workspace, Run (snapshotted at creation), AutodiscoveryRule
- **Migration:** adds `state_mode` column to all three tables
- **API:** workspace create/update accepts `state-mode`, state endpoints guarded
- **Runner dispatch:** `TP_STATE_MODE` env var flows listener → job_template → runner Job
- **Entrypoint:** conditionals skip backend override, state download, backend backstop, state upload
- **UI:** State Mode dropdown on create form + workspace settings, amber banner on State tab
- **Autodiscovery:** rules inherit `state_mode` to auto-created workspaces
- **Tests:** all existing tests updated and passing

## Test Plan
- [x] Local test with `backend "local"` — runner preserves backend, skips state ops
- [x] Local test with `backend "s3"` (real GovCloud bucket) — plan + apply succeed, state written to S3
- [x] State endpoint guards — 404 on current-state-version, empty list on state-versions, 409 on upload
- [x] UI — State Mode dropdown visible on create + settings, amber banner on State tab
- [x] Managed mode unchanged — existing workspaces still work normally
- [x] All unit tests passing (1466 passed)

### DIFF
--- a/alembic/versions/a1b2c3d4e5f6_add_state_mode.py
+++ b/alembic/versions/a1b2c3d4e5f6_add_state_mode.py
@@ -1,0 +1,33 @@
+"""Add state_mode to workspaces, runs, and autodiscovery_rules.
+
+Revision ID: a1b2c3d4e5f6
+Revises: 42266b856e8e
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "a1b2c3d4e5f6"
+down_revision = "42266b856e8e"
+
+
+def upgrade() -> None:
+    op.add_column(
+        "workspaces",
+        sa.Column("state_mode", sa.String(20), nullable=False, server_default="managed"),
+    )
+    op.add_column(
+        "runs",
+        sa.Column("state_mode", sa.String(20), nullable=False, server_default="managed"),
+    )
+    op.add_column(
+        "autodiscovery_rules",
+        sa.Column("state_mode", sa.String(20), nullable=False, server_default="managed"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("autodiscovery_rules", "state_mode")
+    op.drop_column("runs", "state_mode")
+    op.drop_column("workspaces", "state_mode")
--- a/docker/runner-entrypoint.sh
+++ b/docker/runner-entrypoint.sh
@@ -302,50 +302,53 @@ if [ -n "$TP_API_URL" ] && [ -n "$TP_RUN_ID" ]; then
             STRIP_DIR="$WORK_DIR/$TP_WORKING_DIR"
         fi
 
-        # Force the runner to use the local backend by writing a terraform
-        # override file (#346). Override files (*_override.tf) are merged
-        # by terraform/tofu with replacement semantics: an override
-        # `terraform { backend "local" {} }` replaces the main config's
-        # `cloud {}` or `backend "x" {}` block, regardless of how the
-        # user wrote it. This avoids any in-place editing of user code.
-        #
-        # The runner MUST execute on the local backend — a remote backend
-        # inside the Job recurses straight back into Terrapod. So our
-        # override has to WIN, unconditionally. Two mechanisms:
-        #
-        #   1. Filename `zzzz_terrapod_backend_override.tf`. Override
-        #      files are merged in lexical order with the LAST file
-        #      winning, so the `zzzz` prefix makes ours win against a
-        #      user's `override.tf` or any realistic `*_override.tf`.
-        #      (We do NOT defer to a user-supplied backend override —
-        #      deferring to a user `cloud {}` / `backend "remote"` would
-        #      hand the runner a remote backend. Ours always wins.)
-        #
-        #   2. The post-init backstop below
-        #      (.terraform/terraform.tfstate backend.type check) — a
-        #      hard guarantee for the residual case of a user file that
-        #      sorts even later than ours.
-        #
-        # If the user did ship their own override file declaring a
-        # backend/cloud block, log it so the override is visible in the
-        # runner log — but still write (and win with) ours.
-        #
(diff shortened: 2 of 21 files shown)


mattrobinsonsre:
@the3venthoriz0n, thanks for this — and apologies for not landing it. I want to be straight with you about why, because the motivation is right and your work here is real engineering, not just a one-liner you're hoping someone would take.

The pain you've identified is the **right** pain to be solving: state migration is the hardest part of adopting any platform layer, and a hard-cutover-or-bust shape blocks teams with hundreds of workspaces. The diff you put together — model + migration + API + runner dispatch + entrypoint + UI + autodiscovery + tests — is a clean implementation of the design you proposed. You weren't fighting the codebase; you understood it.

But I can't take it in this shape, and the reason is more structural than tactical. Terrapod's value proposition is built on the assumption that **Terrapod owns the state**, not just orchestrates around it. A pile of features only work because of that:

- **Cross-workspace `terraform_remote_state` allowlist (#344)** — the producer-controlled grant model is meaningless if Terrapod isn't actually serving the state read. With external state, S3 IAM is the gate, not Terrapod.
- **Unified lock semantics** — the workspace lock and the state-version lock are one mechanism. External-state mode hands the state lock to DynamoDB; Terrapod can't see, coordinate, or surface that.
- **State-diverged backstop** — the runner flags the workspace if a state upload fails after a successful apply (the apply happened but Terrapod doesn't see the result; #339). External-state runs have no such backstop because Terrapod isn't in the state-write path at all.
- **State versioning, history, rollback** — the State tab, the rollback API, the audit trail of state writes — all gone for external workspaces. The UI gets an "this workspace's state isn't here" gap.
- **The #346 post-init backend backstop** — that's the *guarantee* that a runner never executes against a remote backend (it's a hard `exit 1` if init configures anything other than `local`). External mode disables that backstop. The class of bug #346 exists to prevent (a runner silently running against a remote backend) is exactly what we'd be opting back into.
- **The "single execution shape" of Terrapod**. There is one model right now: runners use local backend, state lives in Terrapod, runs go through the artifact path. `state_mode=external` introduces a second model with different semantics for half the features, and every doc, every runbook, every "what should I expect" has to fork to cover both. The support matrix doubles for the lifetime of the feature.

Each of those is fixable in isolation — you could add S3 lock observation, a state-stub for the UI, etc. — but the cumulative weight is a different product. The value of Terrapod being a single, opinionated shape is precisely that everything composes; external state breaks the composition rule.

That said, **the underlying problem is real and shouldn't go unaddressed.** The right shape of compensation, I think, is the [Terrapod migration tool spike (#347)](https://github.com/mattrobinsonsre/terrapod/issues/347), which I've just broadened to explicitly cover Atlantis migration as a first-class source alongside TFE. The relevant part for what you're solving here is design question #3 — the spike now requires picking and documenting **one** of two state strategies for the Atlantis path:

- **(a) Leave state in place** — Terrapod runs the workspace against the user's existing S3+DynamoDB, no backend rewriting. Cleaner cutover but parallels external mode (and so inherits the same Terrapod-loses-state-ownership trade-offs).
- **(b) Migrate state into Terrapod** — one-shot `terraform init -migrate-state` per workspace, then Terrapod owns it. More invasive but converges on Terrapod's single shape.

These are *the same choice your PR makes*, just resolved at migration time (one-shot) rather than runtime (forever). One-shot lets the operator validate the cutover per workspace, roll back if needed, and end up in a known-good shape without the long-tail support matrix.

The Atlantis migration path in #347 is genuinely under-designed right now (I broadened it from a TFE-only spike), and your firsthand experience with Atlantis adoption pain is exactly what the spike needs. If you have appetite for that work — even just contributing to the design discussion before any code — I'd love it. You've clearly already thought through the side-by-side / gradual / rollback shape; that's the input the spike needs.

Closing this PR with thanks. Genuinely sorry it doesn't land — the reasoning isn't "we don't want this kind of contribution"; it's "this specific shape would compound rather than reduce future complexity, and there's a better place to direct the energy."

PR #372, opened 2026-05-26
feat: add VCS-connected policy sets for syncing .rego from git repos

the3venthoriz0n:
feat: add VCS-connected policy sets for syncing .rego from git repos

Policy sets can now source policies from an external git repository
instead of inline Rego only. A new `policy_vcs_poll` scheduled task
checks tracked branches for new commits and syncs .rego files into
the policies table automatically.

- Model: source discriminator (inline/vcs) + VCS fields on PolicySet
- Poller: extracts .rego from tarball, upserts/deletes to match repo
- API: VCS source on create, inline CRUD guarded (409), manual sync endpoint
- UI: source selector on create form, sync status banner, read-only display


mattrobinsonsre (commented):
## Thanks for this! 🎉

This is a useful feature — pointing a policy set at a git repo is exactly how teams maintain Rego at scale, and you've shipped the full surface (model, migration, poller, API, UI). The shape is right, the distributed-scheduler registration in `app.py:103` is correct (you found that without docs, which is nice), and the inline-CRUD guards at `policy_sets.py:414-418` and `462-468` are the right safety pattern. The schema discriminator with `ondelete="SET NULL"` on the VCS connection is also sensible.

A few things to address before this can merge, plus some design points worth discussing.

### Must-fix before merge

1. **`ruff format` fails on three files** — running `make lint` locally:
   ```
   Would reformat: terrapod/api/routers/policy_sets.py
   Would reformat: terrapod/services/policy_vcs_poller.py
   Would reformat: tests/services/test_policy_vcs_poller.py
   ```
   `ruff format services/` should do it. CI's "Python Lint" check enforces format-on-save.

2. **`tarfile.open` inside an async handler** — `_extract_rego_files` at `policy_vcs_poller.py:53-85` is called from `_sync_policy_set`, which runs inside the asyncio event loop. `tarfile.open(...)` + `f.read()` are synchronous I/O — they block the event loop while the archive extracts. The project has a hard rule against sync work inside `async def` after a previous incident where a synchronous `tarfile` operation made the API's `/health` and `/ready` probes time out for ~2 minutes, killing the pod. The fix is to wrap the extract in `asyncio.to_thread`:
   ```python
   rego_files = await asyncio.to_thread(_extract_rego_files, archive, ps.policy_path)
   ```
   Small archives might not bite in practice, but the rule is "if it's sync I/O / CPU on a buffer > ~10 MB, it goes in a thread." `services/terrapod/api/routers/tfe_v2.py` `upload_state_content` has a recent example using the same pattern.

3. **Build failures are environmental, not your code** — the `Build terrapod-*` failures are "installation not allowed to Write organization package" on GHCR. That's a fork-PR permissions limitation, not anything you did. Once a maintainer rebases/merges this will be a non-issue.

### Design questions

4. **Reimplementing VCS provider dispatch** — `_parse_repo_url`, `_get_branch_sha`, `_get_default_branch`, `_download_archive` at the top of `policy_vcs_poller.py` each switch on `conn.provider` and dispatch into `github_service` / `gitlab_service`. There's an existing abstraction at `services/vcs_provider.py` (the `VCSProvider` protocol) that does exactly this dispatch, and the existing `services/vcs_poller.py` and `services/registry_vcs_poller.py` both consume it. Switching to that means a future Bitbucket / Gitea provider only has to be added in one place. Worth grepping `vcs_poller.py` for `vcs_provider.` to see the consumption pattern.

5. **The sync endpoint runs the sync inline** — `POST /policy-sets/{id}/actions/sync` at `policy_sets.py:348-364` calls `_sync_policy_set` directly inside the request handler. A slow VCS provider then blocks the request for tens of seconds, and on the client side it looks like a hang. The project pattern is to enqueue a trigger task (`scheduler.enqueue_trigger`) and return 202 Accepted with the status fields the UI is already going to poll (`vcs_last_synced_at`, `vcs_last_error`). The triggered handler is what runs the sync. `services/notification_dispatcher.py` is a small example of that pattern; the registration sits next to your `register_periodic_task` call in `app.py`.

6. **Importing a `_`-prefixed function across modules** — `policy_sets.py:354` does `from terrapod.services.policy_vcs_poller import _sync_policy_set`. The underscore prefix is a Python convention for "module-private"; importing it across modules ties the two files together implicitly. Two options: (a) promote it to `sync_policy_set` (no underscore) and document it as the public sync entrypoint, or (b) refactor the sync endpoint to enqueue a trigger per #5, in which case the action handler doesn't need to import the internal at all.

7. **Per-cycle commit can lose work on a mid-loop failure** — `policy_vcs_poll_cycle:158-172` loops all VCS policy sets calling `_sync_policy_set` with no per-set commit, then commits once at the end. If sync #5 of 10 triggers a DB error, the in-memory mutations to sets 1–4 are also lost. Per-set commit (`await db.commit()` inside the loop after each `_sync_policy_set`) means partial progress sticks even on a partial failure. The cost is N round-trips instead of 1, but VCS poll cycles are infrequent.

8. **No archive size cap** — `_download_archive` reads the full tarball into memory before extraction. For `.rego` repos the practical size is small (text files), but a pathological/malicious repo could ship a multi-GB archive and OOM the worker. The project's recent state-upload work added a 256 MB cap as a defence-in-depth measure (`api/routers/tfe_v2.py:1717`) — same idea applies here, even if the realistic limit is smaller.

9. **Soft-deleted VCS connections** — `_sync_policy_set:91` handles `conn is None` (the FK SET NULL case), but doesn't account for the case where the VCS connection still exists as a row but is soft-deleted/disabled. A quick check of `services/vcs_poller.py`'s filter (`VCSConnection.deleted_at IS NULL` or similar) would tell us whether that's something to guard against.

### Test coverage

10. **The extraction unit tests are excellent** — `test_policy_vcs_poller.py` is hermetic, covers the right edge cases (subdirs, trailing-slash, non-rego files, empty archive), and the in-memory tarball builder is a clean pattern. Nice work.

11. **No coverage on `_sync_policy_set` itself** — the actual sync flow (idempotency when SHA unchanged, error-write-to-row when branch missing, delete-on-removal when a `.rego` is gone from upstream, partial-failure inside the loop) isn't unit-tested. The extraction tests cover ~30% of the new code by line. Worth adding a few cases against a fake VCS provider (httpx mock or similar — there are examples in `tests/services/test_vcs_poller.py`).

### Smaller observations

12. `policy_vcs_poller.py:161` `PolicySet.enabled == True  # noqa: E712` — SQLAlchemy 2.0 idiom is `.is_(True)` which doesn't need the noqa.
13. `models.py:1597` adds `source` as a free `String(20)`. A CHECK constraint (`source IN ('inline', 'vcs')`) at the DB layer would keep bad values out even if API validation is bypassed (migration tool, manual psql, etc.). The codebase has examples in `alembic/versions/*` — search for `op.create_check_constraint`.
14. `policy_vcs_poller.py:148` `ps.vcs_last_error = str(e)[:500]` truncates aggressively. Python stack traces are routinely longer than 500 chars; 2000 chars (and storing the *tail* with the actual `Traceback (most recent call last):` line via `str(e)[-2000:]`) gives operators a more useful diagnostic.
15. The migration at `c7d8e9f0a1b2:17` uses `server_default="inline"` for `source` — good, that backfills existing rows correctly. Same for the other string columns with `server_default=""`. Solid.

### Out of scope — suggest splitting

16. **`docker/runner-entrypoint.sh`** — wraps `head -c` with an empty-output fallback message when OPA has no stderr. That's a genuine improvement (we've seen ambiguous errors in production where OPA failed silently), but it's unrelated to VCS-sourced policy sets and reviewing it together makes both PRs harder to land. Worth lifting into its own small PR.

17. **`web/src/app/workspaces/[id]/runs/[runId]/page.tsx`** — restyles the policy panel to break out per-policy results visually. Also genuinely better than what's there (the new "passed/failed/errored" indicator with the colored left border reads much more clearly), but again — orthogonal to the VCS-sourced policy set feature. Lifting it lets you ship that improvement without it being gated on the rest of this PR's review.

---

Big-picture: the design is sound and the implementation is ~80% there. The two blockers above (format + async-thread) are mechanical fixes; the design questions are worth a back-and-forth but none of them require throwing work away. Looking forward to seeing the next iteration. Let me know if any of the project-pattern references (`vcs_provider.py`, the scheduler trigger pattern, `asyncio.to_thread`) are unfamiliar and I can sketch concrete examples.

mattrobinsonsre (changes requested):
Thanks for picking this up — VCS-connected policy sets are a real gap, and the overall shape is right (discriminated `source`, `vcs-*` attrs on the JSON:API resource, periodic sync via the distributed scheduler, 409 guards on inline CRUD when source is `vcs`). A handful of things need to land before this is mergeable, and a couple of architectural details to discuss.

## Required changes

1. **go-terrapod SDK update is missing.** CLAUDE.md "API ↔ Consumer Contract" treats this as a hard requirement: every new endpoint and JSON:API attribute must land in `go-terrapod` (typed struct fields + tags, plus the new `actions/sync` method) in the same PR. Without it the contract drifts immediately and the provider + migrate are blind to the new shape.

2. **`docs/api-reference.md` is not updated.** The new `source`, `vcs-connection-id`, `vcs-repo-url`, `vcs-branch`, `vcs-policies-path`, `vcs-last-sync-at`, `vcs-last-commit-sha`, `vcs-last-error`, and the `actions/sync` endpoint all need entries. This is the surface UI + go-terrapod read; release audit catches it later, easier to do now.

3. **No router tests.** Tests exist for `_extract_rego_files` and `_sync_policy_set` (good), but the router layer is uncovered: `create_policy_set` with `source=vcs`, the 409 guards on `update_policy`/`delete_policy` when source is `vcs`, and the `actions/sync` endpoint. Please add at least one of each.

4. **Tarball safety guards.** `tarfile.open(..., mode="r:gz")` over an untrusted archive needs either Python 3.12+ `data_filter` or an explicit guard. The `/`-in-`remainder` check covers nested paths inside the repo prefix but not direct `../etc/passwd` entries or absolute paths. Code currently only `tar.extractfile()`s contents (no disk write), which limits blast radius, but treat untrusted tarballs as hostile and refuse traversal explicitly.

5. **`_MAX_ARCHIVE_BYTES = 256 MB` happens after download.** By the time the guard fires, `download_archive` has already pulled the whole tarball into a `bytes` buffer in memory. On a deliberately-large adversarial repo (or just a hundreds-of-MB legitimate one — that's a real shape we've hit) this OOMs the API replica before we even check. Cap during download (Content-Length pre-check + chunked read with a running counter, abort on overflow), or use the existing sparse-fetch path that `vcs_archive_cache` already wires for run artifact tarballs.

## Architectural discussion

6. **Per-cycle inline loop in `policy_vcs_poll_cycle` head-of-line blocks across policy sets.** One slow repo (network timeout) stalls every later policy set for the per-set network budget. The scheduler already gives you the trigger queue — `enqueue_trigger("policy_vcs_sync_one", policy_set_id)` per set + a per-set trigger handler is the same pattern `vcs_poll` uses for webhook-driven polls, and means slow repos don't starve fast ones. Worth doing now rather than after the first prod page.

## Nits (non-blocking)

- `import uuid` inside `handle_policy_vcs_sync` is unnecessary — already top-level.
- `_MAX_ARCHIVE_BYTES` is referenced before its definition in the file.
- `f"vcs-{uuid}"` formatting for `vcs-connection-id` then `removeprefix("vcs-")` on parse works but isn't a convention other VCS routers use; consider the same shape as `vcs_connections.py`.

Happy to re-review once 1–5 are addressed (6 is open for discussion). Image-push CI failures are the expected ghcr.io fork-PR denial; the substantive checks (Python Test, SAST, CodeQL, Helm Lint) are green.

the3venthoriz0n:
- Add go-terrapod SDK: PolicySet struct with VCS fields, CRUD methods,
  and SyncPolicySet action (API-consumer contract requirement)
- Add docs/api-reference.md entries for VCS attributes and actions/sync
- Add router tests: create with source=vcs, 409 guards on inline CRUD,
  sync endpoint with enqueue_trigger assertion
- Add tarball path traversal guards (reject absolute paths, .. via normpath)
- Add _download_archive dispatcher using correct provider method names
  (github_service.download_repo_archive, not download_archive)
- Use explicit dispatchers for all provider calls, matching vcs_poller.py
  pattern (github_service method names differ from VCSProvider protocol)
- Refactor policy_vcs_poll_cycle into fan-out dispatcher: enqueues one
  trigger per set, preventing head-of-line blocking
- Fix nits: _MAX_ARCHIVE_BYTES before usage, uuid at module level,
  remove dead normpath substring check"

mattrobinsonsre (commented):
## Re-review — this is close 🎉

Huge progress since the last round. Going through the prior blockers and design points:

**Fixed / addressed:**
- ✅ **`tarfile` off the event loop** — `_extract_rego_files` now runs via `asyncio.to_thread` (poller L171), and the archive read-back is threaded too.
- ✅ **Size cap enforced during download** — switched to the provider stream-to-file path with a running counter that aborts at 256 MB before the bytes ever hit memory. This was the OOM concern; nicely done.
- ✅ **Path-traversal guard** — explicit reject of absolute paths + `..` + a `posixpath.normpath` check before the prefix match.
- ✅ **Trigger fan-out instead of an inline loop** — `policy_vcs_poll_cycle` now enqueues one `policy_vcs_sync` trigger per set (with a dedup key), and `handle_policy_vcs_sync` does the work. One slow repo no longer head-of-line-blocks the rest, and the per-cycle-commit data-loss window is gone.
- ✅ **`actions/sync` returns 202** and enqueues the trigger rather than running the sync inline in the request.
- ✅ **No more cross-module `_`-prefixed import** in the action handler.
- ✅ **go-terrapod SDK** — `policy_sets.go` + tests added (contract requirement).
- ✅ **`docs/api-reference.md`** — new attributes + `actions/sync` documented.
- ✅ **Router tests** — `test_policy_sets_vcs.py` covers create-with-`source=vcs`, the 409/422 guards, and the sync action.
- ✅ **DB CHECK constraint** `source IN ('inline','vcs')` in the migration.

**One thing left before merge:**
- ❌ **`ruff format`** — CI's *Python Lint* is red:
  ```
  Would reformat: terrapod/api/routers/policy_sets.py
  Would reformat: terrapod/services/policy_vcs_poller.py
  Would reformat: tests/api/test_policy_sets_vcs.py
  Would reformat: tests/services/test_policy_vcs_poller.py
  ```
  `ruff format services/` then re-commit. That's the only gating code issue.

**Non-blocking (your call):**
- The poller still dispatches GitHub/GitLab directly via `_parse_repo_url` / `_get_branch_sha` rather than the `VCSProvider` protocol in `services/vcs_provider.py`. Not a blocker — just means a future provider has to be added here as well as there. Fine to leave for a follow-up.

The `Build terrapod-*` failures are the fork-PR GHCR-push permission limitation, not your code — they'll resolve on a maintainer merge. Once `ruff format` is in and Python Lint goes green, this is good to go.

the3venthoriz0n:
Oops sorry, the formatting should be fixed moving forward! I addressed the VCSProvider protocol issue as well.

mattrobinsonsre (approved):
Approved. Ruff format clean (Python Lint green), and the bonus VCS provider dispatch refactor — extracting `parse_repo_url` / `get_branch_sha` / `get_default_branch` / `download_archive` into `vcs_provider.py` and turning the existing `vcs_poller.py` helpers into thin delegates — is exactly the maintainability win I was hoping for as a follow-up. Net 51-line trim on the existing poller, dispatch logic preserved verbatim. Merging.

PR #383, opened 2026-05-27
fix(runner): report meaningful error when OPA fails with empty stderr

the3venthoriz0n:
When OPA evaluation fails with exit code != 0 but produces no stderr
output, the runner now reports 'OPA evaluation failed with exit code X
(no stderr output)' instead of an empty string. Previously this left
policy evaluations with no visible error message in the UI.

### DIFF
--- a/docker/runner-entrypoint.sh
+++ b/docker/runner-entrypoint.sh
@@ -303,7 +303,11 @@ tp_evaluate_policies() {
             fi
 
             if [ "$_opa_exit" != "0" ]; then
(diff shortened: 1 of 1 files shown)


mattrobinsonsre (approved):
Nice tiny defensive fix — empty stderr was leaving a confusingly blank error string in the JSON payload. The placeholder substitution before `jq -Rs` is exactly right. Image-push CI failures are the expected ghcr.io fork-PR denial; all substantive checks (Python Test, Lint, SAST, CodeQL, Helm Lint) are green. Thanks!

PR #384, opened 2026-05-27
feat(ui): verbose policy evaluation display with per-policy pass/fail

the3venthoriz0n:
See PR description


mattrobinsonsre (approved):
Nice cleanup of the policy panel — surfacing per-policy pass/fail/error rather than only the deny list is a real UX improvement. The colour-coded left border per state reads cleanly. Minor nit (non-blocking): the inline `style={{borderColor: ...}}` could be Tailwind `border-l-{colour}` utilities for consistency with the rest of the file; happy to merge as-is and address in a follow-up. Thanks!

PR #385, opened 2026-05-27
feat(migrate): add --workspace flag for direct state migration without atlantis.yaml

the3venthoriz0n:
Atlantis deployments running in autodiscovery mode have no atlantis.yaml
file. The --workspace flag lets operators target an existing Terrapod
workspace directly — the tool detects the backend from HCL in the
source directory and pushes state without needing atlantis.yaml.

Usage:
  terrapod-migrate apply --source=atlantis \\
    --source-dir /path/to/terraform/dir \\
    --workspace my-workspace \\
    --target https://terrapod.example.com \\
    --apply

The workspace must already exist in Terrapod (created via autodiscovery,
UI, or API). This is the state-push-only path for teams that manage
workspace creation separately from state migration.

Adds ReadStateFromDir as a public entry point on the atlantis package
for callers that don't need a full Source struct.

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -6,8 +6,10 @@ platforms are supported:
 
 - **HCP Terraform / Terraform Enterprise (TFE)** — one TFE organization
   maps to one Terrapod deployment (Terrapod is single-org).
-- **Atlantis** — `atlantis.yaml` v3 schema; one or more repos map to
-  Terrapod workspaces or autodiscovery rules.
+- **Atlantis** — `atlantis.yaml` v3 schema, or autodiscovery mode
+  (no `atlantis.yaml` — use `--workspace` for direct per-workspace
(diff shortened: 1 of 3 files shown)


mattrobinsonsre (approved):
Nice extension to terrapod-migrate for the autodiscovery-style Atlantis case where there isn't a canonical `atlantis.yaml` to enumerate workspaces from. `loadDirectWorkspacePlan` + pre-seeded `WorkspaceRecord` is the right shape — reuses the existing writer path rather than forking it. Two small things to clean up in a follow-up (non-blocking for this merge): (1) `docs/migration.md` lines 211 and 228 lost their leading-two-space indent (likely an editor reflow); (2) the new `ReadStateFromDir` and `loadDirectWorkspacePlan` paths don't have unit tests yet. Image-push CI failures are the expected ghcr.io fork-PR denial. Thanks!

Issue #231, opened 2026-04-23
API event-loop stalls: blocking calls on async request paths


mattrobinsonsre:
## Summary

The API's `async def` request handlers contain several calls that block the asyncio event loop — sync CPU-heavy work, pure-Python crypto, sync archive/tarfile operations — none of them wrapped in `asyncio.to_thread()` / `run_in_executor()`. On uvicorn we run **one event loop per worker process**, so any blocking call inside an `async def` parks every other coroutine on that replica (including `/health` and `/ready` probes) until it yields. When a blocker runs longer than the liveness timeout, the pod is SIGKILLed.

One of these has already caused a confirmed ~2-minute production stall + liveness-induced pod restart. Others haven't been observed yet but will fire under the right traffic. This issue catalogs all of them and proposes fixes.

## Confirmed incident

Pod `terrapod-api-5d4957d469-7lzpc` on v0.17.0, mgmt cluster:

```
11:00:13.429Z  Registry VCS poll complete
11:00:15       last /health 200 OK
── ~2 minute gap, nothing in the log ──
11:02:20.726Z  Run created   run_id=019dba01... workspace=sls-stg-us1
11:02:20.747Z  Run transitioned  pending → queued
11:02:20.8     SIGTERM arrives (liveness failed)
```

```
Restart Count:   1
Warning Unhealthy 111s (x3 over 2m21s) Liveness probe failed: context deadline exceeded
Warning Unhealthy  40s (x14 over 2m33s) Readiness probe failed: context deadline exceeded
```

The unblocking event at 11:02:20 was an external `POST /api/v2/runs` with `vcs-ref=04894337...` on `markupai/scalable-language-servers`. The connection was dropped (no response to caller) because SIGKILL arrived mid-request.

### HIGH — on the request path, unbounded size

All of these sit on endpoints that can be hit with multi-MB payloads (state files, archive tarballs) and have no thread-pool offload.

| # | Location | Blocker | Triggered by |
|---|---|---|---|
| H1 | `services/archive_utils.py:22-37` — `strip_archive_top_level_dir()` | `tarfile.open(r:gz)` + member-by-member repack | `POST /api/v2/runs` with `vcs-ref` → `run_service._fetch_vcs_config` (the confirmed 2-min stall) |
| H2 | `api/routers/state_management.py:155` + `:226` | `hashlib.md5(state_bytes)` / `hashlib.md5(body)` on state upload path | State rollback, manual state upload |
[...]

PR #391, opened 2026-05-28
feat: display module inputs/outputs in registry

the3venthoriz0n:
Extract Terraform variable/output declarations from module tarballs
at publish time using python-hcl2. Show them in a collapsible card
on the module detail page. Versions section also now collapsible.

### DIFF
--- a/services/terrapod/services/module_hcl_parser.py
+++ b/services/terrapod/services/module_hcl_parser.py
@@ -97,14 +97,18 @@ def _extract_variables(parsed: dict) -> list[dict]:
     for var_block in parsed.get("variable", []):
         for var_name, var_config in var_block.items():
             has_default = "default" in var_config
-            variables.append({
-                "name": var_name,
-                "type": _type_to_string(var_config.get("type")),
-                "description": var_config.get("description", ""),
-                "default": _serialize_default(var_config.get("default")) if has_default else None,
-                "required": not has_default,
-                "sensitive": bool(var_config.get("sensitive", False)),
-            })
+            variables.append(
+                {
+                    "name": var_name,
+                    "type": _type_to_string(var_config.get("type")),
+                    "description": var_config.get("description", ""),
+                    "default": _serialize_default(var_config.get("default"))
+                    if has_default
+                    else None,
+                    "required": not has_default,
+                    "sensitive": bool(var_config.get("sensitive", False)),
+                }
+            )
     return variables
 
 
@@ -113,9 +117,11 @@ def _extract_outputs(parsed: dict) -> list[dict]:
     outputs = []
     for out_block in parsed.get("output", []):
         for out_name, out_config in out_block.items():
-            outputs.append({
-                "name": out_name,
-                "description": out_config.get("description", ""),
-                "sensitive": bool(out_config.get("sensitive", False)),
-            })
+            outputs.append(
+                {
+                    "name": out_name,
+                    "description": out_config.get("description", ""),
+                    "sensitive": bool(out_config.get("sensitive", False)),
+                }
+            )
     return outputs
--- a/services/terrapod/services/module_hcl_parser.py
+++ b/services/terrapod/services/module_hcl_parser.py
@@ -1,0 +1,111 @@
+"""Extract Terraform variable and output declarations from module tarballs."""
+
+import io
+import json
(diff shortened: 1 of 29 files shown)


services/terrapod/services/registry_module_service.py
```diff
@@ -176,6 +176,19 @@ async def upload_module_tarball(
...
 
     mod_version.upload_status = "uploaded"
+
+    from terrapod.config import settings
+
+    if settings.registry.module_interface.enabled:
+        from terrapod.services.module_hcl_parser import extract_module_interface
+
+        try:
+            interface = extract_module_interface(data)
```
mattrobinsonsre: **Blocking.** This runs `tarfile.open` + `hcl2.loads` synchronously inside `async def upload_module_tarball`, so it blocks the event loop — and therefore the health/readiness probes — for the full parse. A large tarball can trip the 3s liveness timeout and get the pod SIGKILLed (the #231 failure mode).

```python
import asyncio
...
interface = await asyncio.to_thread(extract_module_interface, data)
```

`extract_module_interface` is self-contained and returns a plain dict, so the hand-off is clean.

services/terrapod/services/registry_vcs_poller.py
```diff
@@ -209,6 +209,17 @@ async def _poll_module(db: AsyncSession, storage, module: RegistryModule) -> Non
...
         archive_bytes = await strip_archive_top_level_dir_async(archive_bytes)
 
+        from terrapod.config import settings
+
+        interface = None
+        if settings.registry.module_interface.enabled:
+            from terrapod.services.module_hcl_parser import extract_module_interface
+
+            try:
+                interface = extract_module_interface(archive_bytes)
```
mattrobinsonsre: Same blocking issue as the upload path:

```python
interface = await asyncio.to_thread(extract_module_interface, archive_bytes)
```

This runs inside the VCS poll cycle (a scheduler task on the event loop), so a slow parse here stalls every other coroutine on the replica too.

services/terrapod/services/module_hcl_parser.py
```diff
@@ -0,0 +1,111 @@
...
+            if not member.isfile():
+                continue
+            if "/" in member.name:
+                continue
+            if not member.name.endswith(".tf"):
+                continue
+            f = tar.extractfile(member)
+            if f is None:
+                continue
+            contents.append(f.read().decode("utf-8", errors="replace"))
```
mattrobinsonsre, line 62: Non-blocking: `f.read()` pulls each root `.tf` fully into memory with no size bound. Legit modules are tiny, but a hostile/garbage tarball could ship a multi-GB `variables.tf` and OOM the worker. A per-file cap (read up to a few MB, skip + log if larger) would harden this. Low priority — it only runs at publish time on authenticated uploads — but worth a guard.

mattrobinsonsre (changes requested):
Really nice feature — the surface is complete (migration, dep, parser, config gate, Helm values + schema, docs, and good parser tests). One blocking issue before this can merge, plus two smaller notes, all inline.

**Blocking — sync work in an async handler (CLAUDE.md hard rule).** `extract_module_interface()` does `tarfile.open` + `hcl2.loads`, both synchronous and CPU-bound, and it's called directly (not in a thread) from two `async def` paths. Any blocking/CPU call on the event loop stalls every other coroutine on the replica, including the `/health` and `/ready` probes — if it exceeds the liveness timeout the kubelet SIGKILLs the pod. This already caused a ~2-minute production stall (#231), and module tarballs can be large. Both call sites need `await asyncio.to_thread(extract_module_interface, ...)`.

Everything else looks good — the GET endpoint reads stored columns (no parsing in the request path), and restricting to root-level `.tf` files is a sensible scope.

the3venthoriz0n:
One more small change here: 


Return a JSON Schema object for each input variable's type alongside
the human-readable type string. Enables downstream web applications to
validate module input forms client-side without a server round-trip.
Removes the intermediate type_raw field — consumers get either the
display string (type) or the machine-consumable schema (type_schema).

mattrobinsonsre (approved):
LGTM —  wrap at both call sites, 5 MB per-file cap added, ruff format clean. The bonus  work is well-scoped and high quality. Merging.

PR #398, opened 2026-05-29
feat: add workspace grouping by path with collapsible tree view

the3venthoriz0n:
Adds a Group by dropdown to the workspace list that organizes
workspaces into a collapsible tree by VCS repo and directory path.
VCS workspaces nest under their repo basename; local workspaces
group separately. Collapse/expand state and grouping preference
persist in localStorage across navigation."

### DIFF
--- a/web/src/app/workspaces/page.tsx
+++ b/web/src/app/workspaces/page.tsx
@@ -14,6 +14,7 @@ import { apiFetch } from '@/lib/api'
 import { useSortable } from '@/lib/use-sortable'
 import { useWorkspaceListEvents } from '@/lib/use-workspace-list-events'
 import {
+  type ParsedFilter,
   hasLabelTerm,
   hasStatusTerm,
   matchWorkspace,
@@ -90,8 +91,8 @@ function WorkspaceGroupRows({
   collapsedGroups: Set<string>
   toggleGroup: (path: string) => void
   pathPrefix: string
-  resolveStatus: (ws: any) => any
-  parsedFilter: any
+  resolveStatus: (ws: Workspace) => { def: { label: string; color: string; pillClass?: string; filter?: string } | null; runId: string | null }
+  parsedFilter: ParsedFilter
   setFilterInput: (v: string) => void
   badgeColors: Record<string, string>
   depth?: number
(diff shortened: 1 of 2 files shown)


web/src/app/workspaces/page.tsx
```diff
@@ -59,12 +68,176 @@ interface Workspace {
...
+              </td>
+            </tr>
+
+            {/* Workspace rows */}
+            {!isCollapsed && group.workspaces.map(item => {
+              const ws = item.workspace as Workspace
+              const { def, runId } = resolveStatus(ws)
+              const dir = ws.attributes['working-directory'] || ''
+              const lastSegment = dir ? dir.split('/').pop() || dir : ''
+              const displayName = lastSegment || item.name
```
mattrobinsonsre, line 138: The grouped rows use `dir.split('/').pop()` as the link text instead of `workspace.name`. That changes what operators see: a workspace named `webapp-staging` with `working-directory = "apps/web"` will render as `web` in the tree. The name is the canonical handle for the workspace everywhere else in the UI (workspace detail page, filter chips, API responses) so swapping it under one render mode is going to be confusing. Suggest keeping `workspace.name` as the visible label and, if the directory context is genuinely useful, showing it as a secondary line or hover tooltip.

mattrobinsonsre (changes requested):
Nice direction — grouping is a real ergonomic gap for any deployment with more than ~20 workspaces, and the implementation is cleanly isolated (one new module, one page change, no API surface). Persisting mode in URL + collapsed paths in localStorage is the right division. One bigger design point, and a couple of smaller ones.

**Generalise the modes.** Right now `'path'` actually does **two** things: it groups by VCS repo first, and then by working-directory tree within each repo. That's the natural view for a **monorepo** deployment. For a polyrepo deployment (most workspaces have an empty `working-directory`), the second axis is mostly noise and exposing it as the only grouping mode feels like overkill. I'd split the mode along the two axes:

```ts
export type GroupMode = 'flat' | 'repo' | 'repo-path'
```

- `'flat'` (was `'none'`) — sorted-but-flat list, current default behaviour
- `'repo'` — group only by repo, flat within (polyrepo-friendly)
- `'repo-path'` (was `'path'`) — repo + working-directory tree within each repo, the current `'path'` behaviour kept verbatim

The `'flat'` rename is just for clarity — "none" reads like "no grouping" but the list is still sorted, and "flat" describes what the user actually sees. `'path'` → `'repo-path'` makes the two-axis nature explicit and removes the implication that paths might collapse across repos (which they shouldn't — `terraform/` in two different repos is unrelated). The URL param keeps the new strings.

**Forward compatibility.** The file comment says `WorkspaceGroup` is designed for a future API-driven "project" model. Worth pinning that promise: any future mode (e.g. `'label:team'`) should just add a value here, not require changes to the tree shape or the `WorkspaceGroupRows` renderer.

**Tests.** `workspace-grouping.ts` has non-trivial recursion + several edge cases (leading slash, empty `working-directory`, multi-level nesting, mixed VCS/local). Worth a unit test file alongside it.

See the inline comment for one smaller point.

mattrobinsonsre:
@the3venthoriz0n you still planning to finish this?  I think you may be pleased with the improved functionality of the workspace filter on the newest versions.

mattrobinsonsre:
Closing due to inactivity.  Reach out to re-open.

the3venthoriz0n:
@mattrobinsonsre still planning to finish this! I've been unable to continue the past few weeks but back in the saddle today. Let me deploy the newest version but I still think there's value here if you're open to it.

PR #511, opened 2026-06-15
fix(ai-summary): scope CODE_DIFF tfvars to the var-files a workspace loads

mattrobinsonsre:
The AI plan-summary `CODE_DIFF` diffed **every** `*.tf` / `*.tfvars` between config versions. In a monorepo where one repo holds one var-file per environment, a change to one environment's var-file appeared in **every** workspace's diff. Workspaces that the change does not affect — whose plan reports zero changes — then had the model rate risk off that irrelevant diff: an empty plan came back **medium risk** because another environment's var-file bumped a value the workspace never loads.

## Root cause

`_extract_tf_files_to_dir` (used to build `CODE_DIFF`) extracted all `.tfvars` from the config-version tarball, ignoring which ones the workspace actually loads. terraform/tofu only load a `.tfvars` if it is auto-loaded (`terraform.tfvars` / `*.auto.tfvars`) or passed via `-var-file` (the workspace's `var_files`). Other environments' var-files are never loaded — but were still fed to the model.

## Fix

Scope `.tfvars` in the diff to what the workspace loads:
- `.tf` files always apply → always included (unchanged).
- `.tfvars` → included only if auto-loaded or in the workspace's `var_files`.
- Matching is rooting-robust (exact / path-suffix / basename) since `var_files` is relative to the working directory while the tarball may be repo-rooted.
- Workspaces with **no** declared `var_files` keep the include-all behaviour (no signal to scope by) — no regression for single-config workspaces.
- The follow-up/chat path passes the same `var_files`, so the cacheable prompt prefix stays byte-identical and still cache-hits.

`CODE_CONTEXT` was already `.tf`-only, so it needed no change.

## Effect on the reported case

For a PR that touches `envs/prod-us1.tfvars`, the `stg-us1` workspace (which loads only `envs/stg-us1.tfvars`) now gets an **empty** `CODE_DIFF`, its plan has no changes, and the summary is correctly low-risk — instead of medium-risk off another env's edit.

## Tests

- `_tfvars_is_loaded`: auto-loaded inclusion, empty-`var_files` include-all, declared-match + other-env exclusion, rooting-robust suffix/basename matching.
- `_extract_tf_files_to_dir`: asserts only `main.tf` + the workspace's own var-file are extracted (other envs dropped); and that empty `var_files` keeps all.
- Full `tests/services` + `tests/api` green (1579 passed).

### DIFF
--- a/docs/ai-plan-summary.md
+++ b/docs/ai-plan-summary.md
@@ -283,6 +283,14 @@ the bulk of the request):
   between this run's config version and the previously-applied
   config version. Empty when there's no prior CV (first run, or
   GC'd). Capped at `ai_summary.code_diff_max_bytes` (default 100 KB).
+  `.tfvars` files are **scoped to the var-files this workspace loads**
+  (its `var_files`, plus auto-loaded `terraform.tfvars` / `*.auto.tfvars`).
+  In a monorepo one repo holds one var-file per environment but each
+  workspace loads only its own; without this scoping a change to
+  `envs/prod-us1.tfvars` would appear in the `stg-us1` workspace's diff
+  and inflate its risk even though that workspace's plan has no changes.
+  When a workspace declares no `var_files`, all `.tfvars` are included
+  (no signal to scope by).
 - `CODE_CONTEXT` — concatenated `.tf` files from this run's
   config-version tarball. Capped at
   `ai_summary.code_context_max_bytes` (default 200 KB).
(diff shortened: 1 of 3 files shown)


mattrobinsonsre:
Superseded by #512: grounding risk in PLAN_JSON is more robust than filtering the diff by var_files — the filter approach was fragile (couldn't account for templates/JSON the plan depends on, and added a heuristic). Closing in favour of the prompt-grounded fix.

PR #543, opened 2026-06-25
feat(auth): add PKCE to upstream OIDC authorize and token exchange

Qubut:
## Summary

Terrapod already uses PKCE for browser/CLI login to itself, but the upstream OIDC connector did not send `code_challenge` on authorize or `code_verifier` on token exchange. Some IdPs require S256 PKCE for the authorization code flow even when a client secret is configured (e.g. Pinniped Supervisor).

This change:
- Adds S256 PKCE to `OIDCConnector.build_authorization_request`
- Persists the upstream verifier in `AuthState` and sends it on token exchange in `/auth/callback`
- Adds unit tests in `services/tests/auth/test_oidc.py`

Without upstream PKCE, IdPs that enforce PKCE reject the authorize request and redirect back with `error=invalid_request` (missing `code_challenge`). The Terrapod callback then fails because no authorization `code` is present.

### DIFF
--- a/services/terrapod/api/routers/auth.py
+++ b/services/terrapod/api/routers/auth.py
@@ -175,6 +175,7 @@ async def authorize(
         code_challenge_method=code_challenge_method,
         idp_state=idp_state,
         nonce=auth_request.nonce,
+        idp_code_verifier=auth_request.code_verifier,
     )
     await store_auth_state(auth_state)
 
@@ -360,6 +361,7 @@ async def cli_sso_redirect(
         code_challenge_method=auth_state.code_challenge_method,
         idp_state=idp_state,
         nonce=auth_request.nonce,
+        idp_code_verifier=auth_request.code_verifier,
         credential_type="api_token",
     )
     await store_auth_state(new_state)
@@ -400,6 +402,7 @@ async def callback(
             callback_url=callback_url,
             code=code,
             state=state,
+            code_verifier=auth_state.idp_code_verifier,
         )
     except ValueError as e:
         logger.error("SSO callback failed", provider=auth_state.provider_name, error=str(e))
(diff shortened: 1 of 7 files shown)


mattrobinsonsre (approved):
Thank you for this — and welcome! 🎉 This is a genuinely high-quality first contribution.

You've identified a real interop gap: Terrapod sends PKCE for its *own* login but wasn't sending it *upstream* to the IdP, which breaks against providers that enforce S256 PKCE on the authorization-code flow even with a client secret (Pinniped Supervisor is a perfect concrete example). The implementation is clean, and a few things in particular stood out:

- You threaded the verifier through the **CLI-login path** (`cli_sso_redirect`) too, not just the browser flow — that's easy to miss.
- It's **backward-compatible**: `idp_code_verifier` defaults to `None`, so any in-flight auth states reconstruct fine via `AuthState(**parsed)`, and the token exchange only adds `code_verifier` when one is present.
- Correct S256 derivation, and you added focused unit tests for both the authorize request and the token exchange.

I'm happy to merge this as-is. There are a couple of tiny polish items that are genuinely **ours** to handle in a quick follow-up, so I won't ask you to iterate on them:

- the S256 challenge derivation now overlaps with `_verify_pkce` in `routers/oauth.py` — we'll factor out a shared `s256_challenge()` helper;
- a one-line note in `docs/authentication.md` documenting upstream PKCE.

Really nice work — thank you for making Terrapod better, and I hope to see more contributions from you. 🙏

mattrobinsonsre:
Merging this now — thank you again, and welcome aboard! 🎉

We'll ship a tiny follow-up polish (factor out a shared S256 helper + a docs note for upstream PKCE) and then cut a release shortly, so you'll be able to pull this in from a tagged build rather than building from source. Thanks for the great first contribution. 🙏

PR #611, opened 2026-06-29
feat(registry): verify upstream checksums + signatures (cache + runner) (#607)

### LINKED ISSUE #607: Verify upstream checksums + signatures before caching external providers and binaries (supply-chain integrity)
## Summary

Terrapod's pull-through caches do **not** verify the integrity of artifacts fetched from upstream. Both the binary cache and the provider network-mirror compute the SHA-256 of *whatever bytes they receive* and record it, but neither compares that hash to an authoritative upstream checksum, and neither verifies a GPG/cosign signature. A compromised upstream, a malicious/poisoned mirror, or a successful MITM (TLS reduces but does not eliminate this) would be cached and served as if legitimate.

This is a supply-chain integrity gap. It pairs with the proxy/CA work (#592) and the air-gap issue (#606): controlling *where* artifacts come from is only half the story — Terrapod must also verify *what* it got.

## Current behaviour (verified in code)

**Binary cache — `services/terrapod/services/binary_cache_service.py`**
- `_fetch_and_store_binary` streams the download and returns `(sha256_hex, size)`, which is stored on the `CachedBinary` row. The hash is **recorded, never checked** against the published `SHA256SUMS`.
- terraform/tofu/terragrunt all publish `SHA256SUMS` per release (terraform & tofu additionally publish a GPG/cosign signature over it). None of these are fetched or verified.
- **Highest severity**: the cached binary is the terraform/tofu/terragrunt executable that runs every plan/apply — i.e. arbitrary code execution if tampered.

**Provider cache — `services/terrapod/services/provider_cache_service.py`**
- `fetch_and_cache_single_platform` computes `shasum` from the downloaded archive but the upstream-advertised shasum is **already in hand** (`meta[platform_key]["shasum"]` / `download_info["shasum"]`) and is **never compared**.
- It then computes the `h1:` dirhash from the same unverified bytes and persists it; the runner's lock-extender splices that `h1` into `.terraform.lock.hcl`. So an unverified (potentially tampered) archive is **laundered into a trusted-looking lock entry**, defeating the verification `terraform/tofu init` would otherwise do. The mirror becomes the trust anchor without acting like one.
- The terraform registry download response also exposes `shasums_url`, `shasums_signature_url`, and `signing_keys` — none are verified (Terrapod already has `pgpy` available, used on the provider *publish* path).

### Binary cache
- [ ] Fetch the upstream `SHA256SUMS` for the tool/version and verify the downloaded artifact's sha256 against it **before** storing/serving. Fail closed (clear error) on mismatch.
- [ ] Where a signature exists, verify the `SHA256SUMS` signature against a **pinned** key before trusting it (HashiCorp's GPG release key for terraform; OpenTofu's cosign/GPG signing for tofu; terragrunt's release sums). This defends against a tampered `SHA256SUMS` itself, not just a tampered binary.

### Provider cache
- [ ] Compare the computed archive sha256 against the registry-advertised shasum before caching/serving; fail closed on mismatch. (Low effort — both values are already present.)
- [ ] Verify the provider's `SHA256SUMS` GPG signature against the registry-advertised `signing_keys` (mirror what `terraform init` does), so the mirror is no weaker than a direct install.
- [ ] Only compute/splice `h1` from bytes that have passed verification.

### Cross-cutting
- [ ] Verification is **on by default** and fail-closed. If an escape hatch is provided for genuinely unsigned/internal mirrors, it must be explicit, narrowly scoped, and logged (not a silent global "skip").
- [ ] Pinned keys shipped with the image / configurable; document the trust roots.
- [ ] Surface a clear, actionable error on verification failure (distinguish "checksum mismatch" from "signature invalid" from "no upstream sums available").

## Acceptance criteria
- A tampered binary or provider archive (sha mismatch) is **rejected**, not cached or served, with a clear error.
- A tampered `SHA256SUMS` (bad signature) is rejected where signatures are available.
- Provider archives are verified against the registry shasum + signature before `h1` is computed/spliced into the lock.
- Tests across tiers: unit (verify match/mismatch/bad-signature), services (fetch path fails closed), and a regression test per failure mode. Docs page on supply-chain verification + the trust roots. `llms.txt` / feature catalogue updated.

## Notes
Interacts with #606 (air-gap): in sealed mode, pre-populated artifacts should carry/retain their verified checksums so verification still holds when upstream is unreachable. Strengthens the security posture that the proxy/CA + restricted-network story implies.

mattrobinsonsre:
Closes #607.

Verifies externally-fetched artifacts against the publisher's signature before Terrapod trusts them — at **two layers**, both on-by-default and fail-closed.

## Honest framing
- **Provider archives** → **defense-in-depth**. A committed `.terraform.lock.hcl` already makes `terraform/tofu init` reject a tampered provider; this adds a cache-side check and stops a tampered archive being laundered into a trusted lock entry via the mirror's `h1`.
- **CLI executables** (terraform/tofu/terragrunt) → the **primary, currently-only** publisher-authenticity check. Nothing downstream verifies the binary the runner executes.

This is hardening, not a patch for an exploitable default-config hole (the original issue wording overstated it — see the [reframing comment](https://github.com/mattrobinsonsre/terrapod/issues/607#issuecomment-4830339566)).

## What it does

**Cache-side (API):** on fetch, verify the artifact's SHA-256 against the publisher `SHA256SUMS` and verify the manifest's GPG signature against a **pinned** key, before caching/serving. Fail-closed.
- Binary cache: verify before the DB row exists (the row gates serving); delete the object on failure.
- Provider cache: compare to the registry-advertised shasum + verify the registry SHA256SUMS signature, **before** computing/splicing `h1`.
- Pinned keys (bundled): HashiCorp `34365D9472D7468F`, OpenTofu `0C0AF313E5FD9F80`, Gruntwork `577774ACA847CC49`.

**Runner-side (executables):** the runner re-verifies the binary it's about to run against the signed `SHA256SUMS` with its own pinned key, and logs a visible trust line:
```
✓ verified terraform 1.9.8 (linux/amd64) — SHA-256 matches signed manifest;
  signature valid (pinned key 34365D9472D7468F, via Terrapod cache)
```
Verification material **always comes from the same source as the binary** — Terrapod cache on the cache path (the API persists + serves the signed manifest), upstream on the fallback. Both tf/tofu paths + the terragrunt download are covered. Even a compromised cache can't forge a valid publisher signature.

## Config
- `registry.binary_cache.verify` / `registry.provider_cache.verify` — `off` | `checksum` | `signature` (default **signature**). The runner inherits `binary_cache.verify` via `TP_VERIFY_BINARIES`, so one knob controls API + runner.
- `registry.binary_cache.signing_keys` (default `{}`) — **operator override** for the pinned publisher keys, keyed by tool. Empty → bundled keys. Lets operators rotate a publisher key or trust an internal re-signing mirror **without a Terrapod image rebuild**; propagated to runner Jobs (set at Job-creation from config, not request-time) so API + runner share one trust set. Built-in `.asc` keys remain the fallback.
- `registry.provider_cache.allow_unsigned` (default `false`) — opt-in: in `signature` mode, degrade to a shasum-only check (with warning) when an upstream advertises no signature material (private registries / non-signing mirrors / some community providers) instead of failing closed. Strict stays the default.

## Architecture
- `terrapod/gpg_verify.py` — dependency-free OpenPGP core (parse/verify/load), shared by the API and the lean runner image.
- `services/artifact_verification.py` — API fetch/verify/persist glue (pgpy in `asyncio.to_thread`).
- `runner/phases/binary_verify.py` — runner verification (sync); pgpy + pinned keys baked into the runner image.

## Tests
- `test_artifact_verification.py` (18) — parse, pinned-key load + key IDs, sign/verify/tamper, fail-closed on mismatch/bad-sig/missing-entry/shasum-absent.
- `test_cache_verification_wiring.py` (2) — binary cache deletes object + raises on verify fail; off skips.
- `test_binary_verify.py` (9) — runner fail-closed paths + source-follows-binary URL rules.

## Notes / honest gaps
- Runner image gains `pgpy` (+ cryptography) — a real but modest size bump on the lean runner image; required for OpenPGP verification.
- Helm values↔schema updated for both `verify` keys (renders on values/local/eval).
- **Not yet live-smoked on Tilt**: the runner verification phase does real downloads, so it needs a Tilt run to confirm end-to-end (unit-tested + image COPY/deps verified; the live exec path is the remaining check before a release).


mattrobinsonsre:
**Live Tilt smoke passed** — runner executable verification confirmed end-to-end on the live stack (closes the smoke gap noted in the PR body):
- `tofu 1.12.3` — **cache path** (`✓ verified … via Terrapod cache`, proving the new API persist+serve-sums endpoints) **and** **upstream fallback** (`✓ verified … via upstream`), pinned OpenTofu key 0C0AF313E5FD9F80.
- `terragrunt 1.0.8` — cache path, pinned Gruntwork **ed25519** key 577774ACA847CC49 (bare-binary + EdDSA shape).
Real downloads, real SHA256SUMS+sig fetch, real pgpy verification against pinned keys, visible trust line in the runner log. Ready to merge.

PR #782, opened 2026-07-10
feat: add workspace grouping by repo and path with collapsible tree view

the3venthoriz0n:
## Summary

- Adds a "Group" dropdown to the workspace list with three modes: Flat, Repository (polyrepo-friendly), and Repository + Path (monorepo-friendly)
- Groups workspaces into a collapsible tree by VCS repo and optionally by working-directory path
- Persists grouping mode in URL (`?group=repo`) and localStorage; collapse state in localStorage
- Includes unit tests for the grouping logic (16 tests)

Closes #398

## Addresses prior review feedback

- `GroupMode = 'flat' | 'repo' | 'repo-path'` (generalized modes)
- Uses `workspace.name` as display label (not directory leaf)
- Forward-compatible renderer (new modes only need a builder function)
- Unit test file with edge case coverage

## Test plan

- [ ] Verify grouped view renders correctly with VCS workspaces
- [ ] Verify "Repository" mode groups by repo only (flat within)
- [ ] Verify "Repository + Path" mode nests by working-directory
- [ ] Verify local workspaces appear in a separate "Local" group
- [ ] Verify collapse/expand persists across page navigation
- [ ] Verify column sorting works within grouped view
- [ ] Verify mobile status badges visible in grouped rows
- [ ] `npx tsx --test src/lib/__tests__/**/*.test.ts` — 16 pass


web/src/app/workspaces/page.tsx
```diff
@@ -1037,7 +1331,51 @@
...
+                  collapsedGroups={collapsedGroups}
+                  toggleGroup={toggleGroup}
+                  pathPrefix=""
+                  parsedFilter={parsedFilter}
+                  setFilterInput={setFilterInput}
+                />
+              </tbody>
+            </table>
+          </div>
+=======
```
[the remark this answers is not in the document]
the3venthoriz0n: Fixed

web/src/app/workspaces/page.tsx
```diff
@@ -1037,7 +1331,51 @@
...
           <EmptyState message={t('empty.noMatch')} />
+>>>>>>> main
```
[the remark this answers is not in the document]
the3venthoriz0n: Fixed

web/src/lib/workspace-grouping.ts
```diff
@@ -0,0 +1,131 @@
...
+    const group: WorkspaceGroup<T> = { key, label, workspaces: [], children: [] }
+    for (const ws of repoWs) {
+      insertWorkspace(group, ws, mode === 'repo-path')
+    }
+    sortGroups([group])
+    result.push(group)
+  }
+
+  if (localWorkspaces.length > 0) {
+    const local: WorkspaceGroup<T> = { key: '__local__', label: 'Local', workspaces: [], children: [] }
```
mattrobinsonsre: `'Local'` is a user-visible string, and it reaches the page raw — `{group.label}` renders it directly. So a French or Japanese user gets a group header in English.

It slipped past the guard rather than being missed: `npm run i18n:lint` is an AST check over **JSX literals**, and this is a string in a `.ts` logic module, which it cannot see. Worth knowing for anything similar.

The fix is to keep this module free of display text and let the caller translate. Return a stable, non-display marker — you already have one in `key: '__local__'` — and have the page do `group.key === '__local__' ? t('group.local') : group.label`, with a `group.local` key added to `en.json` and every other offered locale (the completeness gate will fail the build until all of them have it).

That also keeps the grouping logic pure, which is what makes it unit-testable without a rendering context.

web/src/lib/workspace-grouping.ts
```diff
@@ -0,0 +1,131 @@
...
+
+function partitionByRepo<T extends WsConstraint>(workspaces: T[]) {
+  const vcs = workspaces.filter(ws => ws.attributes['vcs-repo-url'])
+  const local = workspaces.filter(ws => !ws.attributes['vcs-repo-url'])
+
+  const byRepo = new Map<string, { label: string; workspaces: T[] }>()
+  for (const ws of vcs) {
+    const url = ws.attributes['vcs-repo-url']!
+    const key = normalizeRepoUrl(url)
+    if (!byRepo.has(key)) byRepo.set(key, { label: repoBasename(url), workspaces: [] })
```
mattrobinsonsre: The group **key** is the full normalised URL, which is right — two repos never merge. But the **label** is just the basename, so `github.com/org-a/infra` and `github.com/org-b/infra` render as two separate groups both titled `infra`, with nothing on screen to tell them apart. On a polyrepo estate — which is the case this mode exists for — that is fairly likely rather than exotic.

A cheap fix that stays readable: keep the basename when it is unique across the result set, and fall back to `owner/name` only for the ones that collide. Something like counting basenames first, then labelling `count > 1 ? lastTwoSegments(url) : repoBasename(url)`.

Worth a test case too — two repos, same basename, different owners, expecting distinct labels.

web/package.json
```diff
@@ -8,6 +8,7 @@
     "start": "next start --port 3000",
     "lint": "eslint src/",
     "type-check": "tsc --noEmit",
+    "test": "tsx --test src/lib/__tests__/**/*.test.ts",
```
mattrobinsonsre: This script is never invoked by CI, so the 16 tests you wrote will not run on this PR or on anything after it. They would pass locally forever and silently stop reflecting the code.

The web lint job (`scripts/lint.sh`) already runs `npm ci && npm run lint && npm run type-check` in a container — adding `&& npm test` there is the smallest change that makes them real, and it puts them behind the same gate as the type-check.

Worth doing in this PR rather than as a follow-up: a test that has never run in CI is not yet a test, and this is the moment it costs one line.

web/src/app/workspaces/page.tsx
```diff
@@ -1,6 +1,6 @@
 'use client'
```
mattrobinsonsre, line 1: This adds a new interactive surface to a page that has a responsive guard, so per AGENTS.md it needs an assertion in `e2e/tests/responsive.spec.ts` in the same PR — at minimum `expectNoHorizontalPageScroll` with a group mode active, since a deep `repo-path` tree is exactly the shape that pushes a table sideways on a phone.

You have already done the hard part in the implementation (`data-testid="ws-row-status-mobile"` and the `lg:hidden` badges), so this is about pinning that behaviour so a future change cannot quietly undo it.

One behaviour worth covering while you are there: the collapse state lives in localStorage but the mode lives in the URL, so a shared `?group=repo-path` link opens with the *recipient's* collapse state. That is a defensible choice — just make sure it is the one you intended, because the alternative (collapse state in the URL too) makes links fully reproducible at the cost of a longer URL.

mattrobinsonsre (commented):
Thanks for this, and sorry it sat — that's on us, not you. It's a careful piece of work: the mobile status badges are preserved in the grouped rows, the URL is the source of truth for the mode with localStorage only as a fallback, and every locale catalog is filled including the novelty ones. Those are exactly the three things that usually get missed here, so this needed less from me than I expected.

Four things inline. Only the first two would block; the other two are judgement calls.

The PR is also **conflicting with `main`** now — a fair amount has landed on the workspaces page since 13 July (mobile card rendering, status badge extraction). Worth rebasing before doing anything else, since it may resolve some of this for you.

On the test plan checkboxes: don't feel obliged to tick them manually. If the responsive assertion goes into `e2e/tests/responsive.spec.ts`, CI does that work for you on every future change, which is the point of having it there.

the3venthoriz0n:
Thanks for the comments, I'll try to get to this asap

the3venthoriz0n:
@mattrobinsonsre let me know how this one looks, sorry for the wait!

mattrobinsonsre:
Merged — thank you for this, and for the patience while it sat.

It is a genuinely well-built contribution. A few things stood out on review:

The label disambiguation in `labelForKeys` is the nicest part: widening the suffix only for the keys that actually collide, so a lone repository stays a bare basename and only the ambiguous ones grow to `owner/name` or the full `host/owner/name`. Working off the normalised key rather than the first-seen URL is the right call too — it makes the label deterministic across case, trailing slashes and `.git`, which is the sort of thing that bites much later otherwise.

You also did the parts contributors usually miss without being asked: all 32 locale catalogs complete rather than a partial fill, the novelty locales genuinely transformed instead of copied from English, a responsive assertion in the phone-viewport suite, and a unit test file wired into `scripts/lint.sh` so it actually runs in CI rather than only locally. That last one improves the repo beyond this feature.

The reason it lands as-is: it is strictly opt-in. `parseGroupParam` falls back to `flat` for anything unrecognised, and rendering is `groupMode !== 'flat' ? grouped : <the existing table>` — so an operator who never opens the dropdown sees exactly what they saw before. That was the condition for taking it, and the design already met it.

I merged `main` into your branch to bring it up to date (it had gone behind during the v1.6.0 release) and verified the whole set locally on the merged result: tsc, eslint, `test:unit`, both i18n gates across all 32 catalogs, and a full production build. CI is green.

Anything further I spot will go in a follow-up PR rather than holding this one up — you have waited long enough for it, and the feature stands on its own.

mattrobinsonsre:
To be concrete about when you can actually use this: it ships in **2.0, later this month**.

It is on `main` now rather than waiting on a branch, so it is in every build from here, and 2.0 is the release it goes out in.

PR #1217, opened 2026-08-01
chore: probe advisory-token permissions (throwaway)

mattrobinsonsre:
Throwaway probe. Answers one question I asserted without testing: can a workflow's `GITHUB_TOKEN` create a repository security advisory?

Non-destructive — POSTs an invalid payload, so **403 = no permission**, **422 = permission fine, payload rejected**. Nothing gets created either way.

Will be deleted once we have the answer.

### DIFF
--- a/.github/workflows/probe-advisory-token.yml
+++ b/.github/workflows/probe-advisory-token.yml
@@ -70,6 +70,34 @@ jobs:
             *)   echo "VERDICT: $code — read the body." ;;
           esac
 
+      # Now the real question: does the PAT carry the right permission?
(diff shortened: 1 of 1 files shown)


mattrobinsonsre:
Done its job. Answers recorded:

- `GITHUB_TOKEN` + `security-events: write` → **403** on both an invalid *and* a valid payload. It cannot create advisories.
- `SECURITY_ADVISORY_AUTH` PAT → **422** ("missing required keys: summary, description, vulnerabilities") — past auth, permission correct, nothing created.

Also gave us the required payload schema for free: `summary`, `description`, `vulnerabilities` — not `severity`, which is what I'd guessed.

Issue #1468, opened 2026-09-04
ci: cut wall-clock — every run ends with one job alone at the tail (13m main / 21m PR)


mattrobinsonsre:
## Summary

CI wall-clock is 13 min on `main` and up to 21 min on PRs. Both numbers are
dominated by **a single job running alone at the tail of the run**, and it is a
different job in each case. Runner slots are not the scarce resource for most of
the run — for the last several minutes of every run, 19 of the 20 available slots
sit idle.

This issue records the measurements and the four fixes they point at. They are
independent; #1 is a one-line change.

## Method

Per-job `created_at` / `started_at` / `completed_at` pulled from
`GET /repos/{o}/{r}/actions/runs/{id}/jobs`, so queue wait (`started - created`)
is separated from execution time (`completed - started`).

Two reference runs:

- `main`: [33801697811](https://github.com/mattrobinsonsre/terrapod/actions/runs/33801697811) — 789s (13m09s)
- PR: [33850078845](https://github.com/mattrobinsonsre/terrapod/actions/runs/33850078845) — 1055s (17m35s; 21 min as reported by the UI including queue)

Observed concurrency ceiling: **exactly 20** (hit twice in each run) — the free
public-repo standard-runner limit.

## Measurement 1 — `main` is bounded by the integration shard

Concurrency profile, `main` run (each `#` = one running job):

```
t(s)   running  queued
  20      17      13   #################
  80      20       1   ####################
 140      11       0   ###########
 240       3       0   ###
 260       2      21   ##          <- amd64 builds land, 21 jobs go eligible at once
 320      14       5   ##############
 460       4       0   ####
 520       2       0   ##
 540       1       0   #           <- everything else is done
 ...
 720       1       0   #
 740       0       6
 780       0       1               <- CI
```

Milestones:

| t (s) | event |
|---|---|
| 11 | `Prepare` done |
| 70 | `Build terrapod-query (amd64)` done → image builds become eligible |
| 260 | last amd64 build (`terrapod-api`, 180s) done |
| 466 | last E2E shard done |
| 526 | `E2E Report` done — **the entire image path is finished here** |
| 736 | `Python Test (integration)` done (715s, started t=21) |
| 789 | `CI` |

**From t≈540 to t=736 — 3m16s — exactly one job is running.** The
`build-images-amd64 → e2e-test` path completes 4½ minutes before the run does.

Worth noting explicitly since it came up: `e2e-test` already declares
[...]

PR #1471, opened 2026-09-04
ci(1468): stop the merge gate waiting on a job it ignores

mattrobinsonsre:
Fix 1 of #1468. One line (plus a comment explaining why it must stay absent).

`ci-pass` listed `e2e-report` in `needs:` while deliberately omitting it from the `results=()` array that decides pass/fail — so the required status check **waited** for a job whose result could not change the answer. On a PR run that is **5m35s** with one job running and nineteen of twenty slots idle, because the report pulls a ~2GB Playwright image and runs `npm ci` from scratch just to merge eight blob files into an HTML page.

## This does not weaken the E2E gate

| job | what it does | in the verdict? |
|---|---|---|
| `e2e-test` (×8 shards) | actually runs Playwright | **yes** — in `needs` *and* `results=()`, unchanged |
| `e2e-report` | download blobs → `playwright merge-reports` → upload HTML | no — and never was |

A failing E2E test still fails CI. Observed directly today: when `E2E Tests (4/8)` failed on #1455, both that check **and** `CI` went red. That path is untouched.

The tradeoff, stated plainly: if `e2e-report` fails you lose the merged HTML report and read the failure in the shard logs instead.

## Why the exclusion exists

On v1.5.3 a GitHub artifact-storage failure — no Terrapod code involved, artifacts intact per the API, `download-artifact` exhausting its own retries — failed an otherwise correct release, and no rerun could clear it. Keeping the `needs:` edge reintroduced most of that exposure as delay rather than failure.

## Verified before changing

- `Cleanup Tag` does not reference `e2e-report`; line 3042 was the only edge.
- Branch protection requires **only** the `CI` check, so this edge was the whole block.
- YAML still parses; `ci-pass` needs 44 jobs, `e2e-test` retained.

Expected: the `CI` check goes green at t≈720 instead of t=1055 on the reference run. The workflow run itself still finishes the report.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3039,7 +3039,17 @@ jobs:
       - build-images-arm64
       - scan-images
       - e2e-test
-      - e2e-report
+      # `e2e-report` is deliberately absent, and must stay absent. It is already
+      # excluded from the `results=()` verdict below (see the reasoning on the
+      # job itself: an artifact-storage failure with no Terrapod code involved
+      # failed a correct release on v1.5.3). Listing it here anyway made
+      # `ci-pass` WAIT for a job whose result it then ignored — 5m35s of a PR
+      # run with one job on the machine and nineteen slots idle, because the
+      # report merges eight Playwright blobs behind a ~2GB image pull.
+      #
+      # The workflow run still finishes it; the required status check no longer
+      # waits for it. Re-adding this line reintroduces the stall silently, since
+      # nothing about the verdict changes (#1468).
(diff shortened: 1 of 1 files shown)


mattrobinsonsre:
Superseded by #1474, which carries this change unmodified alongside the other tail fix.

Combined because both are workflow-topology edits in different regions of `ci.yml`, verifiable only by CI, and revertable as one — so splitting them bought nothing except serial `update-branch` churn, since branch protection requires up-to-date branches.

PR #1472, opened 2026-09-04
ci(1468): scan both arches per image instead of one job each

mattrobinsonsre:
Fix 3a of #1468. Frees five concurrency slots at the moment the critical path needs them.

`scan-images` was a 5-image × 2-arch matrix — **ten jobs** that all became eligible at the same instant as the eight E2E shards, two `helm-eval-boot` jobs and `oci-conformance`. Twenty-one jobs for twenty slots.

The E2E shards lost **27s, 44s and 57s** of queue wait to it, and the shard that waited 44s was the straggler gating `E2E Report` — so this contention was on the critical path, not in slack.

Now five jobs, each scanning both arches. Same images, same Trivy settings, same findings. Each takes roughly twice as long (~52–172s vs 26–86s), comfortably inside the slack before `create-manifests`.

## The one real hazard, handled

The SARIF `category` must stay distinct per image **and** arch — code scanning keys results on it, so collapsing to a single category per image would have had one arch silently overwrite the other's findings. That is why the arch steps are duplicated rather than looped, and it is asserted in review: categories render as `trivy-${{ matrix.image }}-amd64` and `-arm64`.

## Why not `needs: [e2e-test]`

That would also work *today*, but only because `create-manifests` is gated behind `python-test`. Once the integration shard is split (Fix 2), that slack disappears and deferred scans become the new critical path. Consolidation also cuts runner-minutes; deferral doesn't.

## Verified

- YAML parses; matrix is `image` only; 10 → 5 jobs.
- SARIF categories unique.
- Every other reference is by job **id** (`needs:`, `needs.scan-images.result`), so the wiring the release checklist requires — `create-manifests`, `ci-pass` needs + `results=()`, `Cleanup Tag` needs + failure check — is untouched by the job rename. Only `CI` is a required status check.

Real verification is the CI run on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2100,7 +2100,23 @@ jobs:
   # is why each entry is written to be read by someone outside the
   # project.
   scan-images:
-    name: Scan ${{ matrix.image }} (${{ matrix.arch }})
+    # One job per IMAGE, scanning both arches, rather than one per image+arch.
+    #
+    # As 5x2 this was ten jobs that all became eligible at the same instant as
+    # the eight E2E shards, two helm-eval-boot jobs and oci-conformance — 21 jobs
+    # for 20 concurrent slots. The E2E shards lost 27s, 44s and 57s of queue wait
+    # to it, and the shard that waited 44s was the straggler gating E2E Report,
+    # so the contention sat directly on the critical path (#1468).
+    #
+    # Five jobs scan the same images with the same settings and free five slots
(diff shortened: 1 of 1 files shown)


mattrobinsonsre:
Superseded by #1474, which carries this change unmodified alongside the other tail fix.

Combined because both are workflow-topology edits in different regions of `ci.yml`, verifiable only by CI, and revertable as one — so splitting them bought nothing except serial `update-branch` churn, since branch protection requires up-to-date branches.

PR #1473, opened 2026-09-04
ci(1468): split the integration suite across three runners

mattrobinsonsre:
Fix 2 of #1468. `Python Test (integration)` ran 361 tests serially and was the tail of the run whenever it drew a slow runner — on one reference run it finished at t=736s with everything else done by t=540s, leaving one job on the machine and nineteen of twenty slots idle for over three minutes.

## Why count-balancing is sufficient

Measured across five CI runs by deriving per-test durations from Actions log timestamps: **per-test cost is uniform at 0.58–1.07 s/test**. The work is fixture setup and teardown, which every test pays equally, so a file's duration is its test count times a constant. There are no slow tests to isolate — the four heaviest files are simply the four with the most tests.

## Why three

Roughly **100s of every job is fixed cost** — checkout, image, compose stack, collection — measured at 95s and 112s from job start to first test.

| shards | fast run | slow run |
|---|---|---|
| 1 (today) | 319s | 697s |
| 2 | 210s | 400s |
| **3** | **173s** | **300s** |
| 4 | 155s | 251s |

A fourth shard buys 20–50s of wall-clock for another ~100s of runner time.

## No list to maintain

The split happens in `pytest_collection_modifyitems`, so it sees the **true collected set** — a new test file is assigned automatically. That is deliberate: a hand-maintained path list is the hazard the `unit` slice already warns about, where an unlisted file silently never runs.

For the same reason the planner's tests live in `tests/meta/` — a **directory** added to the unit slice — rather than as another root-level file, which would have grown exactly that list.

## The xdist objection is narrowed, not dismissed

It stands *within* a job: xdist workers share one database and race to create the same tables. It does not apply *across* runners, where each shard gets its own compose stack — which is already how `unit` and `services-api` work.

## Verification

The planner is pure and tested alone — 17 tests: totality (every file in exactly one shard, for 1…29 shards), determinism against input order, balance tolerances, and refusal of nonsense. It sits outside `tests/integration/` because that package's autouse fixtures need Postgres, which would have made an infrastructure test require a database.

End-to-end via collection, which needs no database: the shards select **120 / 121 / 120** tests across 9/10/9 files, and their union is identical to the unsharded **361**.

Worth flagging: an earlier version of that check compared 0 against 0 and reported success, because `addopts = "-v"` overrode `-q` and my grep matched nothing. It now asserts a non-trivial count before claiming a match.

Honest caveat: I could not run the integration suite itself locally (the container store is out of disk — separate issue), so this PR's CI run is its first real execution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -470,23 +470,43 @@ jobs:
           # shards. Local bench (Apple M-series; CI ~50-60% of this):
           #   unit         90s serial → 13s parallel (1495 → 492 tests)
           #   services-api 138s serial → 30s parallel
-          # The integration shard stays serial — its
+          # The integration shards stay serial WITHIN a job: the
           # session-scoped Postgres table-creation fixture races
-          # under xdist workers (each worker would try to create
-          # the same tables) and the suite is small enough that
-          # serial is fine.
+          # under xdist workers, which share one database. That
+          # objection does not apply ACROSS runners — each shard
+          # gets its own compose stack, exactly as the two slices
+          # above already do — so the split is runner-level.
           - name: unit
             # Root-level tests/*.py are NOT auto-collected by the directory
             # shards below — each must be listed here explicitly, or it silently
             # never runs in CI. All of these are pure-unit (no Postgres/Redis/S3).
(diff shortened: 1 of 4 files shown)


mattrobinsonsre:
Folded into #1474, which now carries all three #1468 fixes as one CI change. The commit is cherry-picked unmodified.

Issue #1584, opened 2026-09-10
feat: VCS module autodiscovery for the module registry (dry-run, submodule detection, pick-and-choose)


the3venthoriz0n:
## Summary

Terrapod can autodiscover **workspaces** by scanning a VCS repo and proposing
what to create, but there is no equivalent for the **module registry**. Adding
modules today is one-at-a-time: create a registry module and point it at a
repo/tag manually. For a VCS host with many module repos (and monorepos of
submodules), this is tedious and error-prone.

I'd like to add **module autodiscovery** that scans a VCS connection/repo(s),
detects Terraform/OpenTofu modules (including submodules in subdirectories),
and **proposes** them for addition to the registry — with a **dry-run/preview**
and **manual pick-and-choose** selection before anything is created. It should
mirror the existing workspace autodiscovery machinery.

I'm happy to submit a PR for this if you'd like.

## Motivation

- Onboarding a VCS org/repo full of modules should be a scan-and-select flow,
  not manual entry per module.
- Monorepos with a root module plus nested submodules (e.g.
  `example-management-groups` with a `modules/create` submodule) should be
  discoverable, each proposed as its own directly-callable registry module.
- Operators want to review before committing: see what *would* be added, then
  choose a subset.

## Prior art to mirror (workspace autodiscovery)

The workspace flow already has almost all the pieces; module autodiscovery
should follow the same shape:

- **Rule model:** `services/terrapod/db/models.py:1288` (`AutodiscoveryRule`) —
  scoped by `vcs_connection_id` + `repo_url` + `branch`, matched by `pattern` /
  `ignore_patterns`, with `name_template`, `enabled`, and an `on_directory_delete`
  policy. Dry-run is **not** a persisted column — it's a transient API/service
  concept.
- **Detection + preview + materialize:**
  `services/terrapod/services/workspace_autodiscovery_service.py` — detection at
  `:121/:131/:147`, side-effect-free **`preview_for_paths` at `:384`** (proposal
  row shape at `:464-471`: `{name, working_directory, collision,
  existing_autodiscovered}`), idempotent create at `:194/:477`.
- **API:** `services/terrapod/api/routers/autodiscovery_rules.py` — preview a
  saved rule `GET .../{id}/preview` (`:663`), preview an **unsaved** prospective
  rule `POST .../preview` (`:527`, built via `_build_transient_rule` `:481`),
  and on-demand scan+materialize `POST .../{id}/scan` (`:701`).
[...]

Issue #1583, opened 2026-09-10
feat: module registry support for submodules (subdirectories within a repo)


the3venthoriz0n:
## Summary

Terrapod's module registry has no concept of a **submodule** — a module that
lives in a subdirectory of a repository, addressed with Terraform/OpenTofu's
`//subdir` "double-slash" source syntax. A registry module is keyed solely by
`(namespace, name, provider)` and backed by a single tarball whose `.tf` files
must sit at the tarball root. As a result, a monorepo that contains a root
module plus one or more nested submodules cannot be published as distinct,
directly-callable modules, and **catalog items cannot wrap a submodule**.

I'd like to add first-class submodule support and am happy to submit a PR.

## Motivating example

A common layout is a root module with nested submodules in a monorepo:

git::ssh://git@<vcs-host>/<group>/example-management-groups//modules/create?ref=v1.0.0

Here `example-management-groups` is the root module and `modules/create` is a
submodule. Consumers already call submodules directly via the `//subdir` +
`?ref=` go-getter syntax. Terrapod should be able to:

1. **Ingest** such a repo and register the submodule(s) it contains.
2. Show submodules as visibly **nested** under their parent module in the UI
   (a module and its submodules displayed together).
3. Allow each submodule to be **called directly**, exactly as a top-level
   module is — its own versions, interface, download, and catalog item.

## Current behavior (why it doesn't work today)

The subdirectory is dropped or ignored at several concrete points:

- **VCS URL parsing discards everything past `owner/repo`.**
  `services/terrapod/services/github_service.py` `parse_repo_url` (~L914–959)
  and `services/terrapod/services/gitlab_service.py` `parse_repo_url`
  (~L819–856) return only `(owner, repo)`. A `//subdir` separator or `?ref=`
  query is silently ignored.

- **The VCS poller stores the whole repo, not a subdirectory.**
  `services/terrapod/services/registry_vcs_poller.py` `_poll_module`
  (L154–302; download at L223, wrapper strip at L234, store at L250) downloads
  the entire repo archive at the tag and strips only the single
  `owner-repo-sha/` wrapper, leaving the **repo root** at the tarball root —
  never a nested subdirectory.

- **Interface extraction reads root `.tf` files only.**
  `services/terrapod/services/module_hcl_parser.py` `_read_root_tf_files`
  (L62–83) explicitly skips any file in a subdirectory
  (`if "/" in member.name: continue`), so a submodule's inputs/outputs are
  never parsed.

[...]

PR #1617, opened 2026-09-14
feat(1584): module autodiscovery rules — find the modules in a repository and register them

mattrobinsonsre:
Towards #1584 (the carry to main closes it)

Module autodiscovery for the registry, built to work like workspace autodiscovery. A **rule** names a repository on a VCS connection. It says which Terraform files count, using a glob `pattern` plus `ignore-patterns` with the same syntax as workspace rules, and how each module is named and published: name template, provider, tag pattern, labels and owner. Each matching file's directory is a module, either the repository root or a submodule (#1583). Examples, tests, fixtures and hidden directories never count, and only `.tf`/`.tf.json` files do.

- **Preview** (saved or unsaved rule) proposes the candidates. It flags ones already registered, name collisions, and a missing provider.
- **Scan** registers them, all or a chosen subset, as VCS-sourced modules with their `subdirectory`, and reports what it skipped. It's idempotent.
- **Automatic registration.** While a rule is enabled, the registry poll registers directories that *appear* on the tracked branch, then polls their tags in the same cycle. Each poll costs one branch-head lookup per rule, and the tree is walked only when the head moved.
  - Saving a rule registers nothing: its first poll records a baseline.
  - Candidates an operator deliberately left out stay unregistered, because everything a scan or poll has seen counts as seen.
  - Pointing a rule at another repository starts it afresh.
- **Nothing is ever deleted.** Deleting a rule leaves its modules; `module_autodiscovery_rule_id` goes NULL.

### Surfaces
- **API:** `/api/terrapod/v1/module-autodiscovery-rules`, with CRUD, `POST /preview` (unsaved), `GET /{id}/preview` and `POST /{id}/scan` (optional `subdirectories`). Platform admin only. Ids are `modrule-…`.
- **go-terrapod:** rule CRUD, both previews, and a scan of everything or a subset.
- **Provider:** the `terrapod_module_autodiscovery_rule` resource. It brings the provider to 26 resources and 10 data sources.
- **MCP:** `terrapod_module_autodiscovery_rule_list` and `_preview` (read-only), and `_scan` (Act, not destructive: it creates registry modules and touches no infrastructure).
- **Web:**
  - **Admin page** at Admin → Module autodiscovery. It lists rules, has a create/edit form with an unsaved preview, and a per-rule preview with checkboxes for **Register selected** or **Register all**. Delete asks for confirmation. It uses real buttons, shows cards on a phone, and has all 32 locales.
  - **Discover panel** on the module list. It now uses the unsaved-rule preview and links to the admin page.
- **Docs:** `registry.md` (Module autodiscovery), `api-reference.md`, `mcp.md`, `terraform-provider.md`, `autodiscovery.md`, `llms.txt`, README, index.

### Also
- **The one-off `POST /registry-modules/discover` is gone.** It was never released, and the unsaved-rule preview does the same job with patterns. Its SDK method and MCP tool went with it.
- **Fixed a real bug in that endpoint on this branch.** `_walk_repo_for_rule` returns three values but it unpacked two, so every real scan returned a 500. Its tests had patched the walk itself; every test here patches the VCS calls beneath it.

### Migration (1.7 line)
`a7c3e91f52d4` adds the `module_autodiscovery_rules` table and `registry_modules.module_autodiscovery_rule_id` (FK, SET NULL). It's expand-only, sits on the 1.7 head `6dee2fd6e9b0`, and moves `alembic_release_heads.json` "1.7" to `a7c3e91f52d4`.

Proven on real Postgres 16:
- **On this line:** empty → head, then down to `6dee2fd6e9b0`, then up again.
- **Carried to main:** in a main worktree with the migration inserted at the 1.7 branch point and main's `cddfa0bf6bfb` re-parented onto it, main has one head. A seeded 1.7-head database upgrades through all 8 main-only migrations with its data intact, and downgrades back to the 1.7 head and up again.

The carry PR will apply exactly that.

### Tests
- **Service** (`test_module_autodiscovery_service.py`): matching, naming, preview flags, registration, the savepoint race, and polling (baseline, unmoved branch, new directory only, truncated tree, one failing rule, inactive connection).
- **API** (`test_module_autodiscovery_rules.py`): CRUD, validation 422s, 409, the admin gate everywhere, both previews, scan all or a subset, an unknown subdirectory, the 413, and the rate-limit label.
- **Poll hook:** the rules pass runs before the tag poll, and a failing pass doesn't stop tag polling.
- **Integration** (real Postgres): a rule delete keeps its modules, a rescan registers nothing new, names are unique per connection, and a concurrent registration is skipped while the rest commit.
- **Go:** go-terrapod, provider and MCP tool tests. Goldens regenerate additively, apart from the unreleased discover surface.
- **E2E:** the admin page (create, preview, register a subset), the Discover panel, and phone-width guards.

**Verification:**
- venv suite: 4112 passed. The 7 failures are the known venv-only `test_vcs_rate_limit` ×5 and `test_api_opa` ×2.
- Integration tests pass against local Postgres and Redis.
- go-terrapod, provider and mcp tests pass.
- tsc, eslint, i18n:check, i18n:lint and `npm run build` pass.
- The docs audit passes.

PR CI is the gate.


the3venthoriz0n:
Thanks so much for the hard work here but want to flag this. 

One requirement  for #1584: autodiscovery needs to scan the entire org, not just a single repository. As designed here a rule names one repo on a VCS connection, which is the right primitive, but the end goal is that new modules are discovered across all repos in the org without hand-registering a rule per repository — otherwise every new repo is a manual step and modules get missed.

Not asking to solve it in this PR; the per-repo rule is a solid foundation. Just capturing org-wide discovery as a required follow-up so it's tracked against #1584 rather than lost. Design left open — whether that's a rule that targets a whole connection, an org-level rule type, or discovery that enumerates repos and applies existing rule patterns per-repo is up to whatever fits the polling model cleanly.

Let me know if you need more elaboration. TLDR; autodiscovery should scan an org for modules (and submodules)

Issue #1648, opened 2026-09-15
Vault file variables: multi-field templates, whole-secret formats and base64 decode


mattrobinsonsre:
Builds on #1619 (Vault-sourced variables delivered as files). Target: v1.7, then main.

#1619 writes one field of a secret to one file. Real credentials need several fields together, from the same read: an AWS credentials file, a kubeconfig, a PEM bundle whose key matches its certificate.

## Scope

Keys added to the `file` object of a Vault reference. They are validated server-side, and rendering happens on the API side from the single per-run read that #1619 already shares:

- **`template`** (at most 16 KiB): a logic-less `{{ name | filter }}` substitution. When a template is given, `field` becomes optional.
  - Dotted lookup into map fields (`{{ creds.key }}`).
  - Filters: `json`, `base64decode`, `trim`, `lines` (joins a list with newlines), `indent N`.
  - Read-only lease metadata as `_lease.ttl`, `_lease.renewable` and `_lease.expires_at`. The lease id is never exposed.
  - It is a single pass and values are never re-scanned, so a secret containing `{{` cannot inject anything.
  - There are no loops, conditionals, functions or environment, file or network access.
  - An unknown name or filter fails the run, naming the key but never a value.
  - `{{ }}` is not HCL interpolation, so no escaping is needed in `terrapod_variable` values.
- **`format`**: `json` writes the whole data object, or the subset in `fields: [...]`; `env` writes dotenv `KEY="escaped"` lines. `format`, `template` and `field` are mutually exclusive.
- **`encoding: base64`** decodes a field before writing it (for example GCP `private_key_data`).
  - In this release the decoded result must be UTF-8. Anything else is refused with a clear error.
  - Binary files need a wire change and are deferred to 2.0.
- **Size caps:** 256 KiB per rendered file, and a new aggregate cap of 768 KiB across all files in a run.

The renderer lives in its own module, with a source-introspection test asserting it imports nothing that reaches the OS, the filesystem or the network.

## Consumers

- go-terrapod: a round-trip test.
- Provider: docs.
- MCP: the `value_source` schema text plus the catalogue golden.
- Web: a template textarea, format and encoding selects, and a fields list, keeping unknown keys, in every locale.
- Docs: `docs/vault.md`, `llms.txt`, `api-reference.md`.

## Tests

- Renderer goldens: AWS INI, kubeconfig, PEM.
- A `{{` inside a value is not re-expanded.
- Unknown names and filters are rejected.
[...]

Issue #1439, opened 2026-08-27
feat: native Vault as a first-class variable value source (pull secrets at run time, bypass sidecar injectors)


the3venthoriz0n:
## Motivation

Today, operators who keep secrets in HashiCorp Vault have to run a Vault Agent sidecar / injector to land those secrets as files on runner pods, then read them in Terraform via `file("/vault/secrets/…")`. That works, but it has real drawbacks:

- **Out-of-band from Terrapod.** Secret delivery is configured on the Kubernetes deployment (pod annotations), not through Terrapod's own variable model — so it's invisible to workspaces, variable sets, and RBAC.
- **Coupled to pod scheduling and scoped per agent-pool, not per workspace.** Sharing a secret across pools means repeating the annotation config for each one.
- **The obvious workaround loses Vault as the source of truth.** You can put the secret in a Terrapod variable set instead — but a static variable means copying the secret into Terrapod's DB and rotating it by hand, defeating the point of Vault.

What's missing is a way to reference a Vault secret from a Terrapod variable and have Terrapod fetch it at run time — keeping Vault authoritative, while gaining everything the variable pipeline already offers (encryption at rest, per-run delivery, variable sets, workspace scoping).

## Proposed design — a `vault` value source

This maps cleanly onto the pattern already established for git module-auth (#1028): a variable's source is resolved into a concrete value at `next_run`, before delivery, so the runner never sees a reference. `git_http_auth` already does exactly this with `source: static` vs `source: vcs_connection`. A Vault source would slot into the same dispatch:

```json
// a variable whose value is fetched from Vault per-run, never stored in Terrapod
{ "source": "vault", "mount": "kvv2", "path": "apps/netbox", "field": "apitoken" }
```

At run time the server (or listener) authenticates to Vault, reads the path, and delivers the resolved value through the existing per-run Kubernetes Secret as either an env or terraform variable — the two delivery shapes that already exist. No sidecar, no annotations, nothing written to Terrapod's database, and Vault stays the single source of truth.

Because it rides the existing variable pipeline, it inherits variable sets (define-once, assign-many), workspace scoping, and precedence for free — the same leverage the git-auth feature got.

## Authenticating to Vault

**Primary proposal: Kubernetes auth** — the API/listener authenticates to Vault using its own Kubernetes ServiceAccount, which fits Terrapod's K8s-native deployment and avoids a long-lived stored token. (AppRole / static token are reasonable alternatives for non-K8s installs and could follow later.)

[...]

PR #1654, opened 2026-09-15
feat(1619): deliver Vault-sourced variables as files

### LINKED ISSUE #1619: feat: file-path materialization for Vault-sourced variables
## Background

The native Vault variable value source (#1439) shipped with env and Terraform variable delivery, covering both static kv-v2 reads and dynamic secret engines. File-path materialization was explicitly scoped out as a stretch goal / fast-follow and remains unimplemented. `docs/vault.md` states this directly under **Limits**:

> **No file materialization yet.** Values are delivered as environment or Terraform variables. A provider that insists on reading a credential from a file path still needs the sidecar.

## Problem

A class of providers and tools read credentials **only from a file path**, not from env vars or tfvars — e.g. CA bundles, token files, and SDK credential files. For any workspace with such a dependency, the native Vault source is not usable: the Vault Agent sidecar / injector must still be run to land the secret as a file on the runner pod.

This is the exact pattern the native Vault source was meant to replace. Without file materialization, adopting the feature for these workspaces is blocked — you end up running both the native source and the sidecar, which defeats the purpose. **File-path materialization is a hard requirement to use this feature in file-based-credential workspaces.**

## Proposed design

Add an optional **file** delivery mode to a Vault-sourced variable that writes the resolved value to a path on the runner filesystem, mirroring what the sidecar's file-drop does. It should:

- Ride the **existing per-run Kubernetes Secret** — the value is mounted/written from that Secret, never placed in the Job spec, env, or a command line.
- Support a **configurable target path** on the runner, with a sane default location.
- Work for **both static and dynamic** engines, identical to current resolution semantics.
- Remain **agent-execution only**, consistent with the current limit (local execution has no resolution point).
- Inherit the same **log-safety guarantees**: value forced `sensitive`, never returned by the API, never logged; only the reference (and the target path) is stored/returned.
- Preserve current **failure semantics** (4xx → errored run; unreachable/5xx → re-queue).

## Acceptance criteria

- A Vault-sourced variable can declare file delivery with a target path.
- The resolved secret is written to that path in the runner before plan/apply; a provider reading `file(...)` sees it.
- Works with static kv-v2 and dynamic engines.
- No value appears in logs, the Job spec, or API responses.
- Documented as a supported delivery mode in `docs/vault.md`, replacing the "No file materialization yet" limit.

## References

- #1439 — native Vault as a variable value source (file materialization listed as fast-follow)
- `docs/vault.md` — current Limits section

mattrobinsonsre:
Towards #1619.

A Vault-sourced variable can now be delivered as a **file**. Add `"file": {"name": "gcp/adc.json"}` to the reference and the variable's value becomes the file's absolute path, for env and terraform variables alike. The secret itself is never in env or tfvars.

## Behaviour
- **Paths.** A relative name lands under `/var/run/terrapod/files/`. A `~/` name lands under the runner's home (`/home/runner`). The name defaults to the variable key.
  - Names are validated on every write path (422).
  - Paths the runner manages are refused: `.ssh`, `.gitconfig`, `.terraformrc` and `.terraform.d`.
  - Collisions are checked after variable-set precedence.
- **Mounting.** Content travels in a new additive `runs/next` attribute, `vault-files`, into the per-run vars Secret and is mounted read-only. The mode is 0440 with an `fsGroup`, 0444 without. Home files use subPath mounts, and an unprivileged init container creates their parent directories as the runner's UID, so tools can write beside them (e.g. `~/.aws/cli/cache`).
- **One Vault read per secret per run.** Variables naming the same secret take their fields from one response, so a dynamic certificate and its key, or an access key and its secret, come from one credential.
- **Map and list fields** are delivered as JSON, not a Python repr.
- **Refused:** file with hcl, and file on a static source. The `file` keys `template`, `format`, `encoding` and `mode` are reserved and refused until #1648.
- **Old listeners.** An older listener ignores `vault-files`, so the variable points at a file that does not exist: the run fails, and the secret goes nowhere else.

## Consumers
- **Web:** a "Deliver as a file" sub-form on the workspace and variable-set pages. Rebuilding a reference starts from the stored object, so unknown keys, `method`, `data` and future `file` keys survive an edit. Strings are in every locale.
- **go-terrapod / provider / MCP:** docs and schema text, plus a regenerated MCP catalogue golden.
- **Provider fix, existing since #1439:** a Vault variable with `sensitive` unset planned an in-place update on every run. `KeepComputedWhenUnchanged` keeps `version_id` and `updated_at` when nothing the operator controls changed.
- **Docs:** `docs/vault.md` gets "Delivering as a file", plus `api-reference.md` and `llms.txt`.

## Tests
- **Backend:** name validation tables, collisions, the size cap, one read per secret, and 4xx → errored / 5xx → requeued in file mode.
  - **No leaks:** the value is absent from the Job spec, env/tfvars, the API response, the listener and API logs, refusal messages, and an old listener's view.
  - **Integration:** kv2 and dynamic, end to end.
  - **Mutation checks:** 11 guards broken one at a time, each caught.
- **Web:** a 60-case reference-preservation table. E2E: create → reload → edit an unrelated field → turn delivery off, on both pages, plus a phone-width guard.
- **Go:** a go-terrapod round-trip; a provider plan → apply → refresh → plan cycle through the protocol server (no drift; a rename is an in-place update); MCP verbatim send.
- **Local results:** venv suite 4372 passed, plus the 9 known venv-only failures. Go builds, vets and tests clean. `i18n:check` and `i18n:lint` pass. PR CI is the gate, including the integration and E2E tiers, which only run there.

A live smoke on the Tilt stack follows before merge: kv2 relative file, `~/` file with a map field, dynamic TOTP, and no value in the Job spec, logs or API.

Carries to main after merge (`Closes #1619` there).


mattrobinsonsre:
Live smoke on the local stack (Tilt, dev Vault, OpenTofu 1.12.6): **passed**. It was one plan-only run with three Vault variables in file mode:

- **A kv2 field as a relative file.** `ca_path` = `/var/run/terrapod/files/certs/ca.pem`, and `sha256(file(...))` matches the secret's hash exactly.
- **A kv2 map field as a `~/` file.** `sa_path` = `/home/runner/.config/gcloud/sa.json`, delivered through the subPath mount. `jsondecode(file(...)).client_email` works, so the map arrived as JSON, not a Python repr.
- **A dynamic engine with the default name.** The TOTP code file holds six digits.

Pod spec:
- The two relative files are in the read-only files volume, mode 0444 (no `fsGroup` here).
- `sa.json` is a subPath mount, and the `home-dirs` init container is present.
- The per-run Secret holds only `vault-file-0/1/2` plus the tfvars.

No secret value appears in the Job spec, the pod spec, the variables API response, or the listener and API logs. The plan log shows only what the config prints on purpose as an output.

Not covered live: an old listener against this API. Tilt redeployed the listener before I could run it. The backend's unit tests cover what an old listener sees in `runs/next`.

PR #1655, opened 2026-09-15
feat(1650): Vault JWT auth with a projected ServiceAccount token, and a custom CA per instance

### LINKED ISSUE #1650: Vault: JWT auth with the API's projected ServiceAccount token, and a custom CA per instance
Target: v1.7, then main.

Kubernetes auth needs Vault to reach back into the cluster (TokenReview). A Vault outside the cluster, whether managed or HCP Vault Dedicated, often cannot. Private CAs are also common in front of Vault.

### `auth.method: jwt`
- The API logs in with its projected ServiceAccount token, and Vault validates it against the cluster's OIDC discovery / JWKS, so there is no reach-back.
- New config: `auth.audience` (default `vault`) and `auth.token_path`.
- The chart projects a `serviceAccountToken` volume (the audience, `expirationSeconds: 600`) on the API Deployment, only when some instance uses `jwt`.
- Kubernetes auth also gains optional `audience` / `token_path`.
- With `namespace: admin` this is the HCP Vault Dedicated topology.

### Custom CA per instance
- `instances[].tls.ca_secret` / `ca_key`. The chart mounts the CA as a file and renders `ca_file` into the ConfigMap, and the client verifies against it.
- Check whether the global `caBundle` / `SSL_CERT_FILE` path is already honoured by the pinned httpx, and document the answer.

Both follow the config-channel contract: `values.yaml`, `values.schema.json`, the ConfigMap render, and `helm-smoke` assertions. The config and Helm contract snapshots grow additively. AppRole, token and namespaces already exist and get docs examples.

## Tests
- A unit test that the JWT login payload is built from `token_path`.
- Helm smoke: the projected volume renders only for `jwt`, `ca_file` renders, and there is no non-secret `TERRAPOD_*` env.
- A live test against a dev Vault's JWT auth pointed at the cluster's OIDC discovery.

mattrobinsonsre:
Towards #1650.

Two additions for a Vault that sits outside the cluster: login with the API's projected ServiceAccount token (`auth.method: jwt`), and a CA per instance.

## JWT auth
- **`auth.method: jwt`.** The API logs in to Vault's JWT auth with a projected ServiceAccount token. Vault validates it against the cluster's OIDC keys, so Vault needs no TokenReview reach-back into the cluster. With `namespace: admin` this is the HCP Vault Dedicated topology.
- **Config:** `auth.audience` (default `vault`) and `auth.token_path` (default `/var/run/secrets/terrapod/vault/<instance>/token`). The mount defaults to `jwt`.
- **Kubernetes auth** gains the same optional `audience` / `token_path`.
- **Token reading:** the token file is read on every login, because the kubelet rotates it. A missing or empty file fails before any request. A refused login names the audience.
- **Chart:** a projected `serviceAccountToken` volume (the audience, `expirationSeconds: 600`) is added only when an instance needs one.

## Custom CA per instance
- **Values:** `instances[].tls.ca_secret` / `ca_key`. The chart mounts the CA and renders `ca_file` into the ConfigMap. The client verifies the login and every read against an `ssl.SSLContext` loaded from it, and reloads it when the file changes.
- **Conflict:** `tls_skip_verify` with `ca_file` is refused at startup.
- **Global `caBundle`:** verified to already cover Vault. httpx 0.28.1 with `verify=True` honours `SSL_CERT_FILE`, which the chart's `caBundle` sets. A test pins this, so an httpx upgrade that changes it fails CI. The details are in `docs/vault.md`.

## Also
- **Docs** gain AppRole, token and namespace examples. All three already existed but had none.

## Contracts
- The config-key snapshot grows by three keys and removes nothing. The Helm values snapshot is unchanged, because the keys sit under the `instances` list.
- A new `helm-smoke` step checks the ConfigMap keys and defaults, the projection and mounts, and that no `TERRAPOD_VAULT_*` env appears. It also checks nothing is projected without a jwt instance or with Vault disabled.

## Tests
- **Unit:** the JWT login payload is built from `token_path` and re-read on each login; the audience default; SSLContext wiring with and without `ca_file`; config validation; and the missing-token negative path.
- **Venv suite:** 4220 passed, plus the 9 known venv-only failures. `helm lint` and template pass on all three profiles. PR CI is the gate.
- **Live:** a JWT login against the dev Vault on the local stack follows before merge.

Carries to main after merge (`Closes #1650` there).


mattrobinsonsre:
Live test on the local stack (Tilt, dev Vault 1.x, k3s): **passed**.

Setup:
- The dev Vault has `auth/jwt`, configured with the cluster's ServiceAccount signing key and the issuer `https://kubernetes.default.svc.cluster.local`.
- Role `terrapod` is bound to audience `vault` and the subject `system:serviceaccount:terrapod:terrapod`.
- The local instance was switched to `method: jwt`.

Results:
- **Chart:** the API pod came up with the projected token at `/var/run/secrets/terrapod/vault/default/token`. Its claims are `aud: [vault]`, the cluster issuer and a 600-second lifetime. The ConfigMap renders `method: jwt`, `audience: vault` and `token_path`.
- **A plan-only run with a kv2 Vault variable reached `planned`.** The output hash matches the secret exactly, and the value does not appear in the plan log.
- **Vault side:** the API's token was issued by `auth/jwt/login` (role `terrapod`), not by Kubernetes auth.

The custom CA was not exercised live, because the dev Vault is plain HTTP. It is covered by the unit tests and the helm-smoke render.

PR #1667, opened 2026-09-15
feat(1663): Vault diagnostics — per-instance status and a reference check that never mints a credential

### LINKED ISSUE #1663: Vault diagnostics: per-instance status and a reference check that never mints a credential
Target: v1.7, then main.

When a Vault-sourced variable does not resolve, an operator today sees only the run's error. `GET /api/terrapod/v1/vault/availability` reports which instances are *configured*, not whether they work. So "why won't my Vault variable resolve?" means reading API logs. This adds diagnostics that answer that without a run, and without ever minting a credential or returning a value.

### 1. Instance status: admin/audit only
`GET /api/terrapod/v1/admin/vault` (JSON:API) returns, per configured instance:
- `sys/health`: reachable, initialized, sealed, standby, version;
- whether Terrapod can log in with its configured method (kubernetes / jwt / approle / token), and the token's remaining TTL;
- whether TLS verification used a per-instance CA, the global bundle or the default store;
- the last resolution error and when it happened, with names only, never a value.

The status is sampled by a periodic scheduler task into Redis, following the `component_status.py` pattern. Opening the page never logs in to every Vault, and the endpoint stays cheap. It is multi-replica safe (rule 11).

### 2. Reference check: variable-write RBAC
`POST /api/terrapod/v1/workspaces/{id}/vault-reference-checks` takes a reference (or a variable id) and reports, without reading or minting the secret:
- whether the reference parses, and its `file` block validates;
- whether the instance is known and the path is inside its `paths` allow-list;
- whether Terrapod's token can read it, checked with `sys/capabilities-self` on the read path. This never mints a dynamic credential.
- for kv-v2 only, the **key names** present, from the metadata or data read. This never returns values, and never touches a dynamic engine.
- for a template or format, whether the fields it names exist (kv-v2 only).

The check is rate-limited, and it grants nothing beyond what creating the variable and running a plan already gives.

## Consumers (API ↔ consumer contract)
- **go-terrapod:** typed methods for both endpoints, with tests.
- **MCP:** `terrapod_vault_status` (Observe) and `terrapod_vault_reference_check` (Observe), plus the catalogue golden.
- **Web:**
  - an admin Vault status panel;
  - a "Check" action on the Vault reference form in the workspace and variable-set variable forms;
  - every locale, including the transformed joke locales;
  - responsive/mobile rules and E2E coverage.
- **Docs:** `docs/vault.md` (Diagnostics), `docs/runbooks.md` ("A Vault variable won't resolve"), `api-reference.md` and `llms.txt`.

## Tests
- **Services-api:**
  - status sampling for each auth method and failure;
  - capabilities-based checks, including denied, allowed and outside the allow-list;
  - kv-v2 key listing never returns values;
  - a dynamic engine is never read;
  - RBAC negative paths;
  - the rate limit.
- **Source introspection:** the reference-check path never calls the secret-read function on a dynamic engine.
- **E2E:** the admin panel, and the form check with a phone-width guard.
- **Live:** against the dev Vault on Tilt, cover a healthy instance, a sealed or unreachable instance, a denied path, and a kv-v2 key listing.

mattrobinsonsre:
Towards #1663.

Answers "why won't my Vault variable resolve?" without a run, and without ever minting a credential or returning a value.

## Instance status: `GET /api/terrapod/v1/admin/vault` (admin/audit)
- One `vault-instance-statuses` entry per configured instance: reachable, initialized, sealed, standby and version (from `sys/health`), `login-ok` for the configured method, token `ttl-seconds`, `tls-trust` (`instance-ca`, `global-bundle`, `default` or `skip-verify`), and `last-error` with its class, message and time. It carries names only, never a value.
- A scheduler task samples it into Redis every 60s, so requests read Redis and never log in inline.
  - The task is registered only when Vault is enabled.
  - A sealed Vault is never logged in to.
  - Before the first sample every field is null, meaning unknown.
  - With Vault off, the endpoint returns `data: []` and `enabled: false`.
- The last resolution error per instance is recorded from resolution as a best-effort write, kept 7 days, and can never fail a claim.

## Reference check: `POST …/workspaces/{id}/vault-reference-checks`
Needs `var:write`. The admin-only `…/varsets/{id}/vault-reference-checks` exists for the variable-set form.
- The body is a reference, or a `variable-id`. Nothing is stored.
- **Checks:** the reference parses and its `file` block validates; the instance is known; the path is inside the allow-list; the path is readable.
- **Readable** comes from `sys/capabilities-self`: `read` for kv-v2, and `update` or `create` for a dynamic POST. `root` grants everything and `deny` wins. The check never reads the secret.
- **kv-v2 only:** the secret's key **names**, and whether a `field`, template names or format `fields` exist. Listing key names also requires `run:plan`. Without it the check still runs, and says why the names are missing.
- **Dynamic engines are never read.** A source-introspection test pins this.
- The check is rate-limited to 20 per minute per user.

## Consumers
- **go-terrapod:** `GetVaultStatus` and `CheckVaultReference`, with tests and the surface golden updated.
- **MCP:** `terrapod_vault_status` and `terrapod_vault_reference_check` (both Observe), with the catalogue updated.
- **Web:**
  - An admin Vault page, one card per instance with labelled status pills. Its nav entry shows only when Vault is enabled, and only to admin or audit users.
  - A "Check" action on the Vault reference form in the workspace and variable-set forms. It checks exactly what would be saved.
  - Strings are in all 32 locales, with the joke locales transformed.
- **Docs:** a Diagnostics section in `vault.md`, a runbook entry ("A Vault variable won't resolve"), `api-reference.md`, `mcp.md`, `index.md` and `llms.txt`.

## Tests
- **Python:** 97 new tests (72 service, 25 router). The broad venv suite passes 4894, apart from the known venv-only set.
- **Snapshots:** route and attribute changes are additive only.
- **Go and web:** go-terrapod and MCP tests pass. Web `tsc`, lint, unit tests, `i18n:check` and `i18n:lint` pass.
- **Mutation checks:** the dynamic-engine guard and the names-only key listing were each broken on its own, and each was caught.
- **E2E:** the admin page, check round-trips, phone-width guards and RBAC negatives are written. They run in CI.

A live check against the dev Vault and a live UI review follow before merge. Carries to main after merge, with `Closes #1663`.


mattrobinsonsre:
Live check against the dev Vault on the local stack: **passed**.

- **Admin status:** Vault 2.0.4, reachable, initialized, unsealed and active. `login-ok` with Kubernetes auth, `ttl-seconds` 1200, `tls-trust` default, no last error. The sample came from the scheduler.
- **kv2 reference:** all five checks pass. `keys` lists the secret's four key names, never a value.
- **Missing field:** `fields-present` fails with `not present: password`.
- **Denied by policy:** with a temporary `deny` rule on the path, `readable` fails, naming the role, the capability granted and the one the read needs. No keys are listed. The policy was restored afterwards.
- **Dynamic engine inside the allow-list (TOTP):** the capability check passes (`read`), with the note `dynamic-not-read`. No keys are listed and no field check runs.
- **Dynamic engine outside the allow-list (database):** refused at `path-allowed`. Vault's database lease count was **0 before and 0 after**, so the check minted nothing.
- **Path outside the allow-list:** refused at `path-allowed`.
- **No leak:** neither the token nor the CA body appears in any status or check response.

Still to come: a live review of the admin Vault page and the Check action in the UI.

PR #1669, opened 2026-09-15
fix(1591): the plan log keeps following a finished phase until the full log arrives

### LINKED ISSUE #1591: fix(ui): the plan log often misses its tail when a run finishes, until the page is refreshed
When an agent run finishes, the plan log in the run view often stops short. The trailing lines are missing: typically the `Plan: …` summary, the entrypoint's closing lines and post-plan policy output. Refreshing the page shows them. It happens far more often than a dropped connection would explain.

## Why (from reading the code; not yet reproduced under a debugger)

The server side was fixed once already. `_serve_log` in `services/terrapod/api/routers/runs.py` serves the stored log when it exists, and the Redis live stream otherwise. It appends the end-of-log marker (ETX) **only** for the stored log, because:

> the runner uploads the authoritative log to storage from its EXIT trap, which fires AFTER it POSTs plan-result and the run transitions to terminal.

So for a short time after a run turns terminal, the log endpoint serves an incomplete Redis snapshot with no ETX. It is designed on the assumption that the client keeps asking until ETX arrives. The client in `web/src/app/workspaces/[id]/runs/[runId]/page.tsx` doesn't, in two ways:

1. **It stops polling at terminal status.** The 2.5s incremental poll (`useEffect` around line 890) runs only while the status is `planning` or `applying`. On the transition, the status effect (line ~870) does one reset fetch. That fetch usually lands in the gap before the EXIT-trap upload, so it gets the snapshot without the tail. Nothing asks again. ETX is stripped when present, but its absence is never acted on.
2. **That one fetch can be dropped entirely.** `loadPlanLog(true)` returns at once when `planFetchLock` is held (line ~915). A poll or a `log_updated` fetch in flight at the moment of the transition, which is when they cluster, swallows the reset with no retry. `loadApplyLog` has the same shape.

A refresh works because the page loads after the upload has landed.

## Fix

- **Keep following a finished phase's log until ETX.** Continue the incremental poll after the phase ends, with a backoff and a bound (for example up to a minute), and stop only on ETX. For a run whose Job never uploaded a log, the server already sends a bare ETX.
- **Never drop a reset.** Queue a reset requested while a fetch is in flight, so it runs as soon as the lock is released.
- **Same treatment for the apply log.**
- **Test it with `node --test`** (`web/tests/`). Pull the fetch/offset/ETX logic out of the page into a small module, and cover:
  - a terminal fetch that gets the snapshot, then the full log with ETX;
  - a reset requested mid-fetch;
  - a phase that ends with no log at all.
- **E2E**: the plan log shows its summary line after completion, with no reload.

mattrobinsonsre:
Towards #1591.

After a run finished, the plan log often lost its tail (the `Plan:` summary, the entrypoint's closing lines and the post-plan policy output) until the page was refreshed.

The server is already correct. For a short time after the phase ends, `_serve_log` serves the live Redis snapshot with no end-of-log marker (ETX), and expects the client to keep asking. The page didn't:
- it stopped polling at terminal status after one reset fetch, which usually landed before the runner uploaded the stored log;
- that one reset was dropped outright if a fetch was already in flight.

## Fix (web only)
- **New `web/src/lib/log-follower.ts`,** used for both the plan log and the apply log:
  - while a phase is streaming, it polls incrementally every 2.5s, as before;
  - once the phase finishes, it **keeps following until ETX**, backing off over 1s, 2s, 4s, 8s, 15s and 30s (one minute in total), then gives up. The refresh button still works.
  - a reset requested mid-fetch is **queued, never dropped**, and plain fetches requested mid-fetch are coalesced into one;
  - after ETX, fetches are no-ops.
- **After the phase finishes, each read starts again from offset 0 and replaces the text.** The live stream is a sliding 500-line, 256 KB tail window, not a prefix of the stored log, so appending the stored log at a stream-derived offset could duplicate or drop lines at the join.
- **Unchanged:** SSE `log_updated` fetches, the loading spinners, the ANSI rendering and `LogPanel`. There are no new UX strings.

## Tests
- **`node --test`:** 13 new tests covering:
  - a snapshot, then the full log with ETX, assembled exactly once;
  - a stored log that is not a prefix of the snapshot;
  - the phase ending mid-fetch;
  - a reset mid-fetch, and coalescing;
  - a bare ETX;
  - the budget bound;
  - failed reads;
  - the apply path;
  - dispose.
- **Mutation checks:** disabling "continue until ETX" fails 6 tests; dropping a mid-fetch reset fails 1.
- **E2E `log-tail.spec.ts`:** the log endpoint and status are stubbed so the first read after the transition has no tail. It asserts the summary appears exactly once with no reload. It runs in CI.
- **Local results:** tsc, lint, `i18n:check` and `i18n:lint` pass. The web unit tests pass 128/128.

A live check on the local stack follows before merge.

Carries to main after merge, with `Closes #1591`.


mattrobinsonsre:
Live check in a real browser on the local stack: **passed**.

The run page was open before the plan finished, and it was never reloaded: a `window` marker survived, and the page's navigation history shows a single `navigate`. The request timing below comes from the browser's own resource-timing record, and the run's timestamps from the API.

| Time | Event |
|---|---|
| 16:26:03.4 | Page loaded; run `planning` |
| 16:26:05 | Run `planned` |
| 16:26:06.2 | Log read: 602 B, the live snapshot, no tail and no end marker |
| 16:26:07.3 | Log read: 602 B, still the snapshot (backing off) |
| 16:26:09.4 | Log read: 602 B, still the snapshot (backing off) |
| 16:26:12.6 | Log read: **4111 B**, the stored log with its end marker. Following stops. |

The page ended with the complete log, including `Changes to Outputs`, `You can apply this plan…` and the closing `tofu apply …` lines. The old code made one reset fetch at the transition (16:26:06.2) and stopped, so the page would have shown only the 602-byte snapshot until someone refreshed.

Issue #401, opened 2026-06-01
AI-generated plan-risk summary between plan and apply


mattrobinsonsre:
## Problem

The most repetitive part of running Terrapod against high-blast-radius infra is reading every line of a long plan to find the actually-risky changes. A real workspace plan can be hundreds of lines of `moved {}` blocks, cascading data-source re-reads, and tag-only metadata churn around a single buried destroy or replacement that you need to spot before confirm-apply.

Operators end up either:
- Reading every line (slow, error-prone, doesn't scale),
- Skim-checking and missing the dangerous diff (the failure mode this issue exists to prevent), or
- Asking another engineer to "tell me this is safe" (which is what this issue would automate).

## Proposal

A pre-apply summarisation hook that consumes the plan JSON, calls an LLM, and surfaces **two outputs** in two places.

**Outputs:**
1. **A human-readable description of what's changing** — written for an operator skim, in their world's vocabulary. e.g. "Brings the `mai-stg-us1-services1` EKS cluster's vault auth mount under Terraform management for the first time. Adds the standard set of policies + auth roles (acrolinx_platform, ai_service, content_cube, reporting_api, reporting_store, terraform, terraform_markupai). Also retires the pulsar_sync resources keyed on pulsar-stg-eu1 (no consumer)."
2. **A risk surface** — bucketed, structured, with confidence per finding.

**Surfaces:**
- **Run detail page** in the Terrapod UI — primary location, full description + risk findings + raw-plan toggle.
- **VCS PR/MR comment** — posted automatically on speculative plan runs whose `run.vcs_pull_request_number` is set. Uses the existing `vcs_comment_dispatch` trigger, same path the existing plan-status + apply-result comments take. Edit-in-place per PR head SHA. Works for both GitHub and GitLab via the existing `VCSProvider` abstraction.

The description gives "what is happening." The risk surface gives "what to worry about." Together they replace the manual walk-through you'd otherwise ask another engineer to do — both in the UI and on the PR before merge.

### Risk-surface framing

**Output is a risk surface, not a safety verdict.** The model never writes "safe to apply" — only "no destructive changes detected", scoped to what the JSON contains. Hallucinated reassurance is worse than no summary at all.

**Buckets:**
- **Destructive** — destroys + `must be replaced` + tainted resources. Always called out.
- **State-only** — `moved {}` blocks, refresh-only re-reads. Surfaced as "no infrastructure change" with a count.
[...]

PR #1756, opened 2026-09-21
feat(1755): the PR comment says what changed, what it costs and what gated it

### LINKED ISSUE #1755: PR status comment reports neither the plan's shape, its cost, nor why a run was blocked
### What the PR status comment reports today

On a pull request, Terrapod posts a per-workspace comment carrying the run status and a link to the run (plus the AI plan summary when enabled). The PR-level status comment in `vcs_status_comment.py` renders a table, but in practice a reviewer rarely sees it carry anything: it reports `queued` / `changes` rather than the plan's shape, and nothing about cost or governance.

That leaves the reviewer without the three things a PR review actually turns on:

1. **How big is this change?** The add/change/destroy counts exist on the run (`resource_additions` and friends) but the comment does not read them — `_plan_summary` has a standing TODO saying the counts aren't available.
2. **What does it cost?** The runner's `cost_estimate.json` already carries `diff` ("the monthly delta this run introduces", `go-terrapod/cost_estimate.go`), but only `total` is cached on the run, so the delta is not reachable without re-reading the artifact.
3. **Was it allowed, and by what?** A run held by a mandatory policy set, an enforced security scan or a post-plan run task shows nothing in the PR. The reviewer has to open the run page to find out why it is blocked — or, worse, not notice that it is.

### Two structural reasons it stays empty

Both are independent of the rendering, and either one alone keeps the table blank:

- **`vcs_poller` only creates a `PRSession` for `apply_then_merge` workspaces.** The status comment is driven off that row, so in the default `merge_then_apply` mode the trigger never fires and the comment is never posted at all — even though the renderer has always had a `merge_then_apply` cell ("will apply on merge").
- **Nothing refreshes the comment after the plan.** The only enqueue on the plan path happens when the poller *creates* the run, before the plan has run. The counts arrive later with `plan-json-output`, the gates resolve in `complete_plan`, and the cost arrives later still with `cost-estimate` — three separate runner requests. A comment rendered at run-creation time therefore reports a `queued` snapshot permanently.

The second is the same race `_enqueue_plan_json_followups` already documents for the AI summariser and the drift classifier ("the runner POSTs plan-result before uploading plan-json-output"), so the pattern for fixing it is established in the codebase.

### Proposal

Make the PR comment answer those three questions:

- Read the persisted resource counts into the Plan column (`+3 ~1 -2`, falling back to the current wording when the counts are null, so older runs are unaffected).
- Cache the cost `diff` alongside the existing `total` (two nullable columns, expand-only) and report it as a monthly delta, distinguishing "priced, no change" from "not estimated".
- Add a per-workspace block listing every gate that can block and how each ruled — mandatory policy sets, enforced scans, the post-plan task stage — in the order `post_plan_hold` evaluates them, so the first failing gate is the one `blocked-by` names. Advisory results are excluded because they cannot hold a run; passing gates are listed, so the comment is an attestation rather than only an alarm.
- Create the `PRSession` in both workflow modes, and refresh the comment as each of the three inputs lands.

Happy to open a PR with this if the approach looks right — it is implemented and tested, including an end-to-end check against a live instance with a mandatory policy set. Two things worth flagging for review: it adds a second Terrapod comment on `merge_then_apply` PRs (the per-workspace comment still carries the AI summary, so suppressing it would lose information), and it introduces a third copy of the run-URL expression that `slack_notify_service.run_url` and `vcs_status_dispatcher` already build, rather than hoisting a shared helper into the notification path.

A3C7:
Closes #1755.

The PR status comment now answers the three questions a review turns on — how big is the change, what does it cost, and was it allowed — for every affected workspace.

```
### Terrapod — all affected workspaces

| Workspace | Plan | Cost Δ | Apply | Mergeable |
|---|---|---|---|---|
| [`prod-network`](…) | +3 ~1 -2 | +412 GBP/mo | not applied | yes |
| [`demo-network`](…) | +3 | +11 GBP/mo | will apply on merge | yes |

▸ prod-network — blocked by policy
    🟢 post-plan tasks — run-task, mandatory
    🔴 prod-guardrails — policy, mandatory
▸ demo-network — all gates passed
    🟢 baseline-tagging — policy, mandatory

*Updated 2026-09-21T10:32:44Z*
```

### What changed

- **Plan counts** read from the columns `plan-json-output` already persists, closing the standing TODO in `_plan_summary`. Null counts keep the old `changes` wording rather than claiming `+0`, so runs from before those columns are unaffected.
- **Cost Δ** is the monthly delta the run introduces, not the projected total — `cost_estimate.json` already carries it as `diff`. Cached in two new nullable columns (expand-only migration, real `downgrade()`, verified both directions against Postgres 16) rather than re-derived per render. Null means "not estimated"; a priced plan with no delta says `no change`.
- **Gate verdicts** per workspace, in `post_plan_hold`'s evaluation order, so the first failing gate is the one `blocked-by` names. Advisory results are excluded — they cannot hold a run — and passing gates are listed, which is what makes the block an attestation rather than only an alarm. An overridden failure reports as passed but stays named, so the override is visible in the PR.
- A workspace whose mandatory gate failed is **no longer offered** a `terrapod apply` that gate would refuse.
- **`PRSession` is created in both workflow modes.** It was apply-then-merge only, so in the default mode the comment was never posted at all, although the renderer has always had a merge-then-apply cell.
- **The comment is refreshed as each input lands.** The counts arrive with `plan-json-output`, the gates resolve in `complete_plan`, the cost arrives with `cost-estimate` — three separate runner requests, measured ~14s apart on a live instance. Each refreshes under its own dedup reason: sharing one key per session lets `enqueue_trigger`'s `SET NX` keep the first enqueue and silently drop every better-informed refresh for 300s. This is the same race `_enqueue_plan_json_followups` documents for the AI summariser and the drift classifier, and the distinct-key remedy the drift re-trigger already uses.
- The second commit closes PR sessions in both modes (widening creation without widening cleanup stranded them at `state='open'`), which required scoping the cancel query first: it matched `vcs_pull_request_number` with no repo or workspace filter, so closing PR #7 in one repository cancelled active runs for PR #7 in every other one.

### Testing

`make test` green: 7568 passed (7540 before this branch). Also verified end to end against a live single-node instance with a mandatory OPA policy set — a real PR rendered `+3`, `no change` and a passing gate, with the workspace name linking to its run. Three of the defects above were found by that live run rather than by the unit tests.

### Two things worth your judgement

1. **A merge-then-apply PR now carries two Terrapod comments** — this table plus the existing per-workspace comment. I left the per-workspace one alone because it carries the AI plan summary, so suppressing it would lose information the table does not have. Folding the summary into the table would be the single-comment alternative, but it reworks a #401 surface and seemed yours to decide.
2. **A third copy of the run-URL expression.** `slack_notify_service.run_url` and `vcs_status_dispatcher` each build it already; I added a local `_run_url` rather than import the Slack service into the VCS renderer or hoist a shared helper into the notification path. Happy to do the hoist if you would prefer it.


A3C7:
Thanks — and apologies for the crossed wires: this PR went up a minute after the issue, so it was already open when you replied. It's #1756, with the two commits described there.

## The run-URL hoist — one question before I write it

Agreed on hoisting rather than adding a third copy. Looking properly, there are **five** copies of the expression, not three:

| Where | Form | No `external_url` |
|---|---|---|
| `notification_service.run_ui_url` | named | `""` |
| `slack_notify_service.run_url` | named | `""` |
| `vcs_status_dispatcher:417` | inline | `""` |
| `module_impact_service:619` | inline | `""` |
| `vcs_status_comment._run_url` (mine) | named | `None` |

All five agree on the semantics — no link when `external_url` is unset — just spelled two ways. One difference that looked like a bug and is not: `run_ui_url` is called with *prefixed* ids (`ws-…`, `run-…`) where the others pass bare UUIDs, so the two produce different URLs. Both resolve, because `api/ids.py` registers those prefixes and `parse_id` tolerates them, so notification links are fine — but it does mean a hoist has to preserve each caller's id form rather than normalise it.

So: **how far do you want this to go?**

1. Literally what you asked — lift `slack_notify_service.run_url` to a shared home, my renderer imports it, the other three copies stay.
2. All five — one helper returning `str | None`; the two named functions stay as thin adapters keeping their `""` contract and their existing call sites (including the prefixed-id convention); the two inline copies and mine call the helper. **No URL changes, no contract changes.**

I lean towards 2, since leaving three copies behind rather defeats the point, but it touches the notification path and `module_impact_service`, which are nothing to do with this PR — your call, and I'll do whichever you prefer. Where should it live? `services/terrapod/services/run_links.py` seemed the obvious home next to the other small focused modules, but say if you'd rather it went elsewhere.

## The second comment on `merge_then_apply`

Taking you up on working it out here. The overlap is total except for one thing: the table carries the run link, the counts, the cost and the gates; the per-workspace comment carries the status word, the run link — and the AI summary, which is the only thing the table does not have. Three shapes:

**A. The per-workspace comment becomes the AI narrative, and only that.** Drop its status line and run link (both already in the table), keep the heading so a multi-workspace PR stays legible, and post it only when there is a summary to show.

```
### Terrapod — prod-network

▸ 🤖 AI summary — 🟡 risk: medium
    Adds three random-provider resources and widens the database subnet
    group to a second AZ. No IAM or security-group changes.
```

**B. Fold the summary into the table's per-workspace block**, so a PR carries exactly one Terrapod comment. Nests a `<details>` inside a `<details>` — the narrative lands two clicks deep — unless the summary becomes a sibling block, at which point the one-comment gain is mostly cosmetic. It also needs the summariser to refresh the table when the summary lands, since it arrives after the plan; that is cheap now, as the refresh machinery this PR adds takes a reason per trigger and `summariser.py:1523` already re-enqueues the per-workspace refresh.

**C. Leave both as they are** — what the PR does today.

**I'd suggest A.** It removes all of the duplication, leaves the #401 surface and its async path intact, and answers the objection you raised: suppressing the per-workspace comment loses the AI summary, so instead let that comment *be* the AI summary. The honest wrinkle is that when the summary is off or not yet ready there is nothing to post, so the comment appears once the summary lands rather than at plan time. If that flicker bothers you, C is a perfectly good resting place and B is the tidiest end state if you want one comment per PR badly enough to rework #401's rendering.

Happy to implement whichever, in this PR or a follow-up.

## Noted on the rest

Understood on minor-not-patch and on this staying clear of the security patches — no expectation of it riding along. And thank you for saying what happened with the nine; generalising the unauthenticated-write finding into a sweep over the whole class is a better fix than the four endpoints we reported, and good to know the signing key is generated at install now rather than warned about.

mattrobinsonsre:
No apology needed for the crossed wires — you were ahead of us, which is the better direction for that to go.

## First, the thing that should not be in this PR

Your second commit fixes this:

```python
active = await db.execute(
    select(Run).where(
        Run.vcs_pull_request_number == sess.pr_number,
        Run.status.notin_(run_service.TERMINAL_STATES),
    )
)
```

No repository filter, no workspace filter. Closing PR #7 in one repo force-cancels every active run numbered 7 everywhere — and low PR numbers collide constantly across repos. I checked: it is on `main`, `release/v1.7` and `release/v1.6`, so it is live for anyone running more than one VCS-connected repository today.

**Please split it into its own PR against `main`.** It is a pure bug fix with no schema change, so unlike the rest of this it is patch material and can be backported to both supported lines. Left here it ships only with a minor, and 1.6 and 1.7 users keep force-cancelling each other's runs in the meantime.

Raise it with whatever detail you have on how you hit it — if you saw it live rather than by reading, that belongs in the issue, because it is the kind of thing people have probably been writing off as flakiness.

## Run-URL hoist — option 2, and thank you for counting properly

Five, not three: that is exactly why it was worth asking. Go with **option 2**, and `services/terrapod/services/run_links.py` is the right home.

Two things to bake in:

- **Preserve each caller's id form rather than normalising**, as you propose. Please put the reason in a comment on the helper — that `run_ui_url` passes prefixed ids and the others pass bare UUIDs, that both resolve because `parse_id` tolerates the prefix, and that normalising would silently change the URLs in existing notifications. Without that written down it reads like an inconsistency and someone tidies it.
- Keep the two named functions as thin adapters with their `""` contract, as you said. The `str | None` at the core is the right shape; `""` is a rendering decision.

A separate commit in this PR is fine — no need for another round trip. It is mechanical and reviewable on its own.

## The second comment — **A**, and it is better than you argued for

`ai_summary.enabled` defaults to **false**, in both the config model and the chart. So under A, a deployment that has not turned AI on posts the per-workspace comment *never* — meaning exactly one Terrapod comment per PR, carrying strictly more than today's does.

That turns your honest wrinkle into the common case working out well: two comments only where someone has opted into AI summaries and therefore wants the narrative. The "comment appears when the summary lands rather than at plan time" caveat then applies only to those deployments, where a summary arriving a few seconds after the table is expected behaviour rather than a flicker.

So: A, in this PR or a follow-up, whichever you prefer.

## On the rest

The live-run finding rate is the part I would underline — three defects that the unit tests did not catch. Worth saying in the PR description that the end-to-end check was where they came from; it makes the case for that being routine rather than optional.

We will review the whole thing properly once the security patches are out; they are mid-flight and I would rather give this your actual attention than a distracted pass.

mattrobinsonsre:
One more thing, which is the part worth hearing first: **structurally this is sound and matches how we already work.**

Managing the PR comment incrementally — post early, refresh in place as each input lands — is the model we intend, not a workaround for the three-request race. So your distinct-dedup-key-per-reason is not just the fix for the `SET NX` swallow, it is the established precedent: `summariser.py:1611` already re-enqueues `vcs_commit_status` under `vcs_status:aisum:{run_id}` for exactly that reason, and the drift re-trigger does the same. You landed on the house pattern by reading the failure correctly rather than by being told.

Which means the shape of this is agreed and the remaining work is additive: a couple more refresh cycles as further inputs land, and some internal machinery so everything you want in the table is reliably surfaceable at the moment the comment renders. None of that changes the design you have proposed.

So treat the two open questions as the only real decisions — the hoist scope, which is option 2, and the comment overlap, which is A — and not as doubt about the approach.

mattrobinsonsre:
A heads-up so you don't get a conflict out of nowhere: I've taken the scoping half of `baa663af5` onto the 1.7 release line.

The cross-repo cancellation is a real bug on its own — `_reconcile_closed_pr_sessions` matching runs on `vcs_pull_request_number` with no repo or workspace filter means closing PR #7 in one repository cancels PR #7's active runs everywhere else, and your point that PR numbers are small integers so collisions are the ordinary case is exactly right. That belongs in a fix release rather than waiting on 2.0, so it's going out with the security patches as v1.7.3.

What I took is only the `select(Run).where(...)` scoping and three tests for it, **authored as you** (`git log` on the branch shows you as author; the message says the diagnosis and the patch are yours). What I deliberately left behind is the other half of that commit — moving `_reconcile_closed_pr_sessions` out of the `apply_then_merge` branch so both modes reach it. That changes what an existing configuration does, which isn't patch material for a release line, so it stays here on your PR where it belongs.

So: **nothing for you to do, and please don't drop it from this PR.** When #1756 lands on main, the scoping hunk will already be there and git will resolve it as an already-applied change; if it does snag, take main's side of that hunk — it's your code. The both-modes relocation, its test, and the `TestPollWorkspacePassesMetadataCacheCorrectly` mock fix all stay yours and unchanged.

Shout if you'd rather I hadn't split it — I can just as easily carry the whole commit and leave your PR to rebase.

A3C7:
Crossed wires again, and this time in a way that needs one decision from you.

**The split already happened** — I'd taken your previous comment as the instruction and done it before your heads-up arrived: #1760 for the bug, #1761 for the fix against `main`. So the scoping hunk is **no longer in this PR**. That matters, because your note assumed it still was:

> When #1756 lands on main, the scoping hunk will already be there and git will resolve it as an already-applied change

It won't be there from this PR. So `main` gets the scoping fix from **#1761 or from your own carry, and it needs one of them**. If #1761 gets closed as a duplicate while your carry goes only to the 1.7 line, `main` ships without a fix that v1.7.3 has — a regression for anyone upgrading off the release line, and against your own rule that a release line's changes are a prefix of `main`'s.

Either resolution works and I have no preference:

1. **Merge #1761 to `main`**, keep your cherry-pick for `release/v1.7`. A main-based PR can't reach the release line anyway, so the two are complementary rather than duplicated.
2. **Close #1761** and carry your own to `main` as well as 1.7. Say the word and I'll close it myself.

#1761 may be the more convenient of the two to take: it carries its own issue, and three tests including one that asserts on the query's **WHERE clause** rather than the rendered statement — `select(Run)` names `workspace_id` in its SELECT columns, so my first attempt at that test passed against the unfixed code. It is verified failing on `main` before the fix:

```
E   assert 'workspace_id' in 'runs.vcs_pull_request_number = :vcs_pull_request_number_1
                              AND (runs.status NOT IN (__[POSTCOMPILE_status_1]))'
```

**What stayed here, exactly as you asked:** the both-modes relocation, its test, and the `TestPollWorkspacePassesMetadataCacheCorrectly` mock fix. I've added a comment on the relocation recording that it depends on the scoping fix landing first — reaching that code from every workspace rather than only apply-then-merge ones is what turns the collision from latent into likely, so **this PR should not merge before the fix is on `main`**.

And no, I'd much rather you had split it. It belongs in a fix release and I had no way to put it there.

---

**Also pushed since your review:** the hoist, as option 2.

`services/terrapod/services/run_links.py` — `run_url()` returning `str | None`, all five call sites migrated, `notification_service.run_ui_url` and `slack_notify_service.run_url` kept as thin adapters with their `""` contract. The id-form reasoning is written into the helper's docstring as you asked: which caller passes prefixed ids, that both resolve because `parse_id` tolerates the prefix, and that normalising would silently rewrite the URL in every notification already sent. `grep` finds no remaining hand-built copy. Tests pin that each adapter still returns `""` without `external_url` and still produces the URL it produced before, so the refactor cannot have moved a link.

Option **A** for the comment overlap is next — and thank you for the `ai_summary.enabled` default, which I had not checked. It turns the wrinkle I flagged into the common case landing well: with AI off, that comment never posts and a PR carries exactly one.

No rush on any of this from here; the security release is the thing that matters this week.

mattrobinsonsre:
You've done everything we asked and then some, so a status update — and a warning about what just landed under you.

**#1761 is merged.** Your scoping fix is on `main`, authored as you, and it also went out on the v1.7 patch line. Splitting it out was the right call and it unblocked a release.

**You took both review suggestions rather than arguing them, and implemented them better than described.** `run_links.py` is the hoist we asked for; `refactor(1755): one place knows what a run's URL looks like` is a better name for it than anything in our comment. And `feat(1755): the per-workspace comment carries only what the table cannot` is option A, which is the one we'd have picked.

**Now the bad news: this PR is conflicting, and it is our fault, not yours.** Two things merged to `main` in the last hour — #1761 (your own fix) and #1757, a set of nine security advisories. The conflict is in exactly one file:

```
services/tests/services/test_run_service.py
```

Nothing else in your twenty-file diff collides. #1757 added a test class to that file and you added one too; it is an adjacent-hunk conflict, not a semantic one.

**Before you rebase, one piece of timing.** A second batch of security fixes is queued for `main` and it touches three files you also touch — `run_service.py`, `slack_notify_service.py`, and `test_run_service.py` again. It is not merged yet and it is behind a 1.7 release and a 1.6 release in the queue, so there is a window, but if you rebase today you may well be doing it twice. Your call entirely: rebase now if you want CI green, or wait and we'll tell you here when `main` has stopped moving.

One thing we should flag since it touches your diff: `run_service.confirm_run` has gained a guard refusing to apply a plan-only run, and a reviewer has since pointed out it does not cover every caller the way its comment claims. If your changes route a run to apply through anything other than `confirm_run`, that is worth a look on your side too — we're fixing our end regardless.

No action needed beyond the rebase whenever suits you. Thanks for the patience with a repo that moved three times under a single PR.

A3C7:
**The rebase is done and verified, and I am holding it unpushed** — taking your second option. No reason to spend your runner minutes twice when you have told me `main` is still moving and the queued batch touches three of the same files. Say the word when it has settled and I will push; if that batch conflicts too, better to resolve both in one pass.

State of it locally, on top of `110c52e8`:

- `make lint` green, services suite 3649 passed. (The one failure is `test_api_opa` again — same upstream OPA outage as the CI red on #1761, not the branch.)
- Verified by inspection rather than by the rebase exiting cleanly: your scoping fix is present, the test class #1757 added is present, and my both-modes change now sits **on top of** the scoping fix, which is the ordering my commit message said it needed.

**One correction to your note:** there were two conflicting files, not one. `test_vcs_poller.py` conflicted as well — #1761 appended a class to it and so did I, at the same end-of-file position. Both were adjacent-hunk, both resolved by keeping both classes; nothing semantic in either.

`test_run_service.py` I rebuilt rather than hand-merged — git had interleaved `TestCompletePlanSurvivesADuplicatePost` and `TestCompletePlanRefreshesThePRComment` into overlapping hunks because the two classes have a similar shape. Taking `main`'s file whole and re-appending my class was safer than resolving marker by marker, and both classes are intact.

**On your `confirm_run` question** — my changes add no apply route. `complete_plan` is a wrapper now, but it delegates to the unchanged body and the refresh it adds is read-only reporting; nothing in this PR moves a run toward apply.

I did look at the wider question you raised, and I would rather not discuss the specifics on a public thread, since the guard is the fix for an advisory that has not been published. I will put what I found on that advisory instead. Short version: the concern is real, it is not in this PR, and it is not the `plan_only` dimension.

mattrobinsonsre:
**`main` has settled — push when you're ready.**

You asked for the word, and rather than just give it I ran your rebase to see what you'd actually hit.

## What moved under you

Exactly one merge since `110c52e8`, the base you rebased onto: **#1767**, the carry of the twelve medium advisories to `main`. Current head is `b3f1aba5`, CI green.

Four of your twenty files are in it:

| File | What #1767 did there |
|---|---|
| `services/terrapod/services/run_service.py` | +35, the plan-only guard in `transition_run` and `confirm_run` |
| `services/terrapod/services/slack_notify_service.py` | +75/-1, the URL redaction rewrite |
| `services/tests/services/test_run_service.py` | +83, tests for the above |
| `llms.txt` | +1 |

## I rebased it, and there is nothing new to resolve

I replayed your branch: first onto `110c52e8` to reproduce where you are, then onto current `main`.

**Step two is conflict-free.** Every one of those four files auto-merged — #1767's hunks sit in different regions of each file from yours (`transition_run` ~886 and `confirm_run` ~1686 against your `complete_plan` ~1190; the redaction block at ~178 against your `run_url` at ~101).

Then I ran the seven affected test files on the rebased tree: **302 passed**.

One honest caveat on that number. Reproducing step one meant resolving *your* two conflicts myself, so my tree is not byte-identical to yours and the 302 is evidence about the merge, not a substitute for your own run. The two results that don't depend on my resolution are the ones that matter: step two conflicts with nothing, and the four shared files merge automatically.

## Your migration needs no re-parenting

`c4d1e7b90a26` declares `down_revision = "e2450c5ecc86"`, and `e2450c5ecc86` is still `main`'s alembic head. #1767 modified four existing migrations but added none, so the chain is untouched.

## On the two conflicts you already resolved

For the record, since you resolved them before we ever discussed them: both were additive collisions in test files, four distinct class names, no semantic overlap.

- `test_run_service.py` — `TestCompletePlanSurvivesADuplicatePost` (ours, from #1757) against your `TestCompletePlanRefreshesThePRComment`
- `test_vcs_poller.py` — `TestClosedPRCancelIsScopedToItsRepository` (your own #1761 fix) against your `TestPRSessionIsCreatedInBothModes`

Worth knowing git splits the first of those into *two* conflict regions with shared context between them, so a naive union produces invalid Python — the class header and body collide separately. If your resolution reads cleanly and your suite is green, you already avoided that.

## Context you'd want, having filed them

The batch that was blocking you is out and public: **v1.7.3** and **v1.6.6** are released, and all twenty-one of your advisories are published, crediting you as finder. GitHub will ask you to accept the credit before your name renders on them.

`GHSA-w67x-7rf5-w65g` is the one you kept cross-referencing as "finding 29" — plan-only runs being applied past a mandatory policy set. It's fixed, in `transition_run` rather than `confirm_run`, because the listener status endpoint passes a caller-chosen `target_status` straight past the latter. That guard is one of the `run_service.py` hunks you're now rebasing over.

Push whenever suits. Nothing else is queued for `main` that touches your files.

mattrobinsonsre:
**Good news first: nothing has moved under you since I gave the word.** `main` is still at `b3f1aba5`, exactly where it was when I wrote on the 22nd — so the rebase you're holding at `667198d9` still applies as verified. Nothing to redo.

**The ask, and I want to be straight about why rather than just chase you.** We're putting together **v1.8.0**, and this PR is the last thing in it. Everything else earmarked for the line has landed:

- #1763 (per-workspace settings reachable from every surface)
- #1764 (per-workspace runner debug mode)
- #1765 / #1766
- #1794–#1798 (the five PR-feedback bugs, in review now)

Your work is a natural fit for that release rather than a patch — it adds two columns and a migration, so it needs a minor either way — which is exactly why I'd rather it went out *with* 1.8.0 than trailed behind it.

**So, whichever suits you:**

1. **Push when you're ready** and I'll drive it through CI and handle the `release/v1.8` backport myself. You don't need to touch the release line at all.
2. **If you'd rather take longer**, say so and I'll cut v1.8.0 without it. It then goes into the next minor on its own merits, with nothing lost and the credit unchanged.

The only outcome I'd like to avoid is waiting on something you weren't planning to get to this week — so a one-liner either way is genuinely more useful to me than a fast push. No rush on the work itself; just on knowing which of the two it is.

And to say it plainly, since this is the second time I've written about timing rather than substance: the delay is on our side of the fence, not yours. You did what we asked, verified it, and held it exactly as agreed.

A3C7:
Pushed — option 1. Head is now `1dc97d70`, rebased onto `b3f1aba5`.

One correction for the record: the held rebase was actually on `110c52e8`, not current `main`, so I replayed it once more onto `b3f1aba5`. As you predicted, that was conflict-free — the same 20 files, and the range-diff against the previous head is clean. The migration `c4d1e7b90a26` is still the single head, on `e2450c5ecc86`. Locally, `make lint` is clean and the full `make test` suite passes (7750 passed).

Happy for you to take the `release/v1.8` backport as offered. Thanks for holding the release for it.

PR #1761, opened 2026-09-21
fix(1760): a closed PR cancels runs only in its own repository

### LINKED ISSUE #1760: Closing a PR force-cancels same-numbered runs in every other repository
Closing a pull request force-cancels every active run carrying the same PR number in **every** repository Terrapod tracks, not just the one whose PR closed.

### Where

`vcs_poller._reconcile_closed_pr_sessions` — the poll-cycle fallback that cleans up after a closed PR. When a session's PR is no longer in the provider's open list, it looks up the runs to cancel:

```python
active = await db.execute(
    select(Run).where(
        Run.vcs_pull_request_number == sess.pr_number,
        Run.status.notin_(run_service.TERMINAL_STATES),
    )
)
for run in active.scalars().all():
    await run_service.cancel_run(db, run, force=True)
```

The predicate is the PR number and the run being non-terminal. Nothing constrains the repository, the VCS connection or the workspace — although the enclosing function has `conn` and `repo` in hand and uses both to find the sessions in the query immediately above.

### Why it bites in practice rather than in theory

PR numbers are per-repository and start at 1, so any Terrapod tracking two repositories has colliding numbers almost immediately. Every low number is shared by every repo.

The cancel is `force=True`, so the victim run stops with nothing that distinguishes it from any other cancellation. From the owning team's side a plan simply died for no reason they can see, on a PR nobody touched. That is the kind of thing people write off as flakiness and retry, which may be why it has not been reported.

Three conditions have to line up, all of them ordinary: two repositories on the same Terrapod, a PR closing in one, and a run in flight on the same number in the other.

### Scope

Present on `main`, `release/v1.7` and `release/v1.6`.

The blast radius today is smaller than the code suggests, because `_reconcile_closed_pr_sessions` is only reached for `apply_then_merge` workspaces — an install that uses only the default `merge_then_apply` never runs it. Anyone with an `apply_then_merge` workspace and more than one VCS-connected repository is exposed.

### How this was found

By reading, not by observing it. I hit it while reviewing my own change in #1756: that PR creates the `PRSession` in both workflow modes, which would make this code reachable for every workspace rather than only `apply_then_merge` ones — turning a latent collision into a likely one. I have no production sighting to offer, and no evidence anyone has been bitten; the "written off as flakiness" framing above is an inference from the symptom, not a report.

### Fix

Constrain the cancel to the workspaces on this connection and repository — the same pivot `vcs_status_comment._collect_rows` already uses:

```python
Run.workspace_id.in_(
    select(Workspace.id).where(
        Workspace.vcs_connection_id == conn.id,
        Workspace.vcs_repo_url.endswith(repo),
    )
)
```

PR to follow. It is a pure bug fix with no schema change, so it should be patch material for both supported lines rather than waiting on the minor that #1756 needs.

A3C7:
Closes #1760.

`_reconcile_closed_pr_sessions` cancelled runs matching the PR number and non-terminal status, with no repository, connection or workspace constraint — so closing PR #7 in one repository force-cancelled every active run numbered 7 in every repository the install tracks. Scoped to the workspaces on this connection and repo, the same pivot `vcs_status_comment._collect_rows` uses for the same question.

Split out of #1756 at your request: pure bug fix, no schema change, so it is patch material for `release/v1.7` and `release/v1.6` rather than waiting on the minor that one needs.

**On provenance, since you asked:** found by reading, not by observing it. I hit it reviewing my own change in #1756, which makes this code reachable for every workspace rather than only `apply_then_merge` ones. I have no production sighting and no evidence anyone has been bitten — the "reads as flakiness" framing in the issue is an inference from the symptom, not a report.

### Testing

Three tests, and the first one earns its place twice over. It asserts on the query's **WHERE clause**, not the rendered statement, because `select(Run)` lists `workspace_id` among its SELECT columns — my first attempt at this test asserted against the whole statement and passed against the unfixed code. Verified failing on `main` before the fix:

```
E   assert 'workspace_id' in 'runs.vcs_pull_request_number = :vcs_pull_request_number_1
                              AND (runs.status NOT IN (__[POSTCOMPILE_status_1]))'
```

The other two pin that the reconciliation still does its job: a closed PR closes its session, an open one is left alone. Full `test_vcs_poller.py` green at 79 passed; `ruff check` clean.


mattrobinsonsre:
Reviewed, and this is the right shape. Merging once CI is green.

**It is already in the 1.7 release, authored as you.** I took the same scoping hunk onto the v1.7.3 patch branch before this PR existed (`9927f804` on `security/a3c7-1.7`, in #1758) — `git log` there shows you as the author. So there is no gap to close on that line and nothing for you to do; this PR is the main-line home for it.

Three things I checked rather than took on trust, since you were explicit about provenance:

- **The WHERE-clause assertion is the load-bearing one**, and your note about the first attempt is right — I removed the scoping and watched it fail with the same output you quoted, then restored it. A test that passes against the unfixed code is worse than no test, and this one doesn't.
- **1.6 is not getting it**, and that is a policy call rather than a judgement on the fix: `SECURITY.md` offers that line security fixes only, and this is a correctness bug. If someone on 1.6 is actually bitten, that changes.
- **One non-blocking observation on the pivot.** `vcs_repo_url.endswith("acme/web")` also matches `https://github.com/notacme/web` — a suffix match will cross an organisation boundary when one org's name ends with another's. It narrows the blast radius from *every repository* to *a suffix collision*, which is the fix working; I mention it only because the same pivot is in `vcs_status_comment.py:158` and `vcs_auto_merge.py:74`, so if it is ever worth tightening it is worth tightening in all three at once rather than here. Not something to change in this PR.

**One housekeeping note:** the squash commit will drop the two generated-with lines from the body — this repo keeps assistant attribution out of its git history as a standing rule, and that applies to my own commits too, so it is not a comment on your workflow. `Closes #1760` and everything else is preserved verbatim.

Thanks for splitting it out — the separation between "the bug, which is patch material" and "the feature it was found inside" is exactly what made it shippable this week.

mattrobinsonsre:
Correction to my housekeeping note above — disregard it. Your PR body merges verbatim, generated-with lines and all.

I had the rule backwards: it governs what the maintainers author, not what a contributor puts on their own work. Your commit, your description, your call.

Everything else in that comment stands.

A3C7:
Thanks — and good to know the 1.7 line is already covered as `9927f804`; that closes the gap I was about to raise, since my split had left the scoping hunk out of #1756 on the assumption it would arrive on `main` from here.

Noted on 1.6 being policy rather than judgement, and no argument from me.

**Your observation about the pivot is right, and chasing it turned up a second defect in the same expression — pointing the other way.**

Yours is a false positive: `vcs_repo_url.endswith("acme/web")` also matches `notacme/web`. Anchoring the match fixes that:

| stored `vcs_repo_url` | `.endswith(repo)` | `.endswith("/" + repo)` |
|---|---|---|
| `https://github.com/acme/web` | ✅ | ✅ |
| `https://github.com/notacme/web` | ⚠️ matches | ✅ does not |

But the same table run over the URL forms `parse_repo_url` documents shows a **false negative**, and anchoring does not help it:

| stored `vcs_repo_url` | `.endswith(repo)` | `.endswith("/" + repo)` |
|---|---|---|
| `https://github.com/acme/web.git` | ❌ | ❌ |
| `git@github.com:acme/web.git` | ❌ | ❌ |
| `https://github.com/acme/web/` | ❌ | ❌ |

`parse_repo_url` accepts all three (`github_service.py:1079-1089`) and strips `.git`, so `PRSession.repo` is `acme/web` — while `tfe_v2.py:1370` and `:1946` store `vcs-repo-url` **verbatim**, with no normalisation. A workspace configured with the `.git` or SSH form therefore never matches its own PR session.

The consequence is worse than the collision, because it is silent and it inverts the message: `_collect_rows` returns no rows, and the PR gets

> _No Terrapod workspaces affected by this PR._

on a PR that does affect one. Polling is unaffected, because that path goes through `parse_repo_url` rather than the suffix match — which is why it would not show up as "VCS is broken", only as a comment that quietly says nothing is happening.

**This is pre-existing, and it is not new in my PR** — but #1756 widens who meets it. That comment only existed for `apply_then_merge` before; it is every PR afterwards, so a `.git`-configured workspace goes from having no table to having a table that says the wrong thing.

I have deliberately **not** changed it, in either open PR. A real fix is either normalising `vcs_repo_url` on write (a migration, and a backfill for existing rows) or matching the candidate forms in SQL, and it wants doing once across all three sites you named — `vcs_status_comment.py:158`, `vcs_auto_merge.py:74`, and the one in #1761 — rather than three times differently. Which shape you want is your call; say the word and I will write it as its own issue and PR, or leave it with you entirely.

The reproduction is four lines if it is useful:

```python
STORED = ["https://github.com/acme/web", "https://github.com/acme/web.git",
          "git@github.com:acme/web.git", "https://github.com/acme/web/"]
for url in STORED:
    print(url, url.endswith("acme/web"), url.endswith("/acme/web"))
```

**Separately, on #1756:** option **A** is implemented and pushed, along with the hoist. The per-workspace comment now carries only the AI narrative where the table is present, and is not posted at all when there is none — so with `ai_summary.enabled` at its default a PR carries exactly one Terrapod comment. The caller checks that the table actually exists rather than assuming it, because a module-impact run carries a PR number belonging to the module's repository while its comment goes to the workspace's, so no session matches and it must keep the full comment.

A3C7:
CI on this PR is red, and it is not the change — it is an upstream outage that has since cleared. Flagging it because you said you would merge on green, so it would otherwise just sit there.

**One root cause, two symptoms.** GitHub's release CDN was returning 500 for the OPA binary:

```
Dockerfile.test:34
  curl -fsSL --retry 6 ... "${BASE}/opa_linux_${ARCH}_static"
  curl: (22) The requested URL returned error: 500     ← ×7, every retry
```

- **Python Lint** never got past the image build, so it failed before running `ruff` at all.
- **Python Test** shards built an image without a working `opa`, and failed on exactly the two tests that need it — `test_policy_engine.py::test_check_rego_accepts_valid` and `test_api_opa.py::TestCheckRegoDegrades::test_still_reports_genuinely_broken_rego` — with **1856 passed** alongside them.

**Evidence it is not this branch**, rather than my say-so:

| run | branch | result |
|---|---|---|
| 14:33 | `security/a3c7-1.6` | ✅ before the window |
| 14:49 | `security/a3c7-1.7` | ✅ before the window |
| 14:55 | `fix/1760-pr-cancel-scope` | ❌ |
| 14:55 | `feat/1755-…` | ❌ |
| 15:00 | **`main`** | ❌ |

`main` went red in the same window on the same infrastructure, which is the test that matters. This branch touches `vcs_poller.py` and one test file — nothing in the image build or in OPA.

**It has cleared.** `opa_linux_amd64_static.sha256` for v1.19.0 returns 200 again as of a few minutes ago, so a re-run should go green with no change to the branch.

I cannot press re-run — pull-only here. Say the word if you would rather I force a fresh run by pushing to the branch instead, though on a PR you have already reviewed I would rather not rewrite it just to retrigger CI.

Incidentally this also explains something I had written off: that same `test_api_opa` test failed intermittently for me locally all afternoon and I had put it down to "environment-dependent". It is — the environment being whether the OPA binary is fetchable. Your CI outage and my local flake are the same thing.

mattrobinsonsre:
Merged. Thank you — genuinely.

Splitting this out of #1756 is what made it shippable: the bug fix is patch material and the feature it was found alongside is not, and keeping them together would have meant either holding the fix or dragging a behaviour change into a release line. You did that on your own initiative after one comment, which saved a round trip.

The test is the part worth calling out. Asserting on the WHERE clause rather than the rendered statement is a distinction most people would not have made, and your note that the first attempt passed against unfixed code is the detail that makes it trustworthy — a regression test nobody has seen fail is a guess. We removed the scoping and watched it fail with the output you quoted before merging.

The provenance note was appreciated too. "Found by reading, not by observing it; no production sighting; the flakiness framing is an inference from the symptom" is exactly the right way to hand over a finding, and it changed how we triaged it rather than just how we felt about it.

This is now on `main` and already on the v1.7 patch line, where it went out authored as you.

